PR #1, opened 2016-11-28
Use u64 instead of i64 for the type for recorded values

marshallpierce:
This will make it easier to port my tests over, since those all use unsigned values, and value has to be non-negative anyway. This diff is mostly pretty mechanical; 95% of it is just making obvious changes to keep the compiler happy.

If you'd prefer I could submit this (and future) PRs to a different (non-master) branch if you'd like to accumulate this stuff elsewhere, though I don't plan on ever making a PR that would leave master in a broken state.

### DIFF
--- a/src/iterators/linear.rs
+++ b/src/iterators/linear.rs
@@ -6,6 +6,7 @@ use iterators::{HistogramIterator, PickyIterator};
 pub struct Iter<'a, T: 'a + Counter> {
     hist: &'a Histogram<T>,
 
+    // > 0
     valueUnitsPerBucket: u64,
     currentStepHighestValueReportingLevel: u64,
     currentStepLowestValueReportingLevel: u64,
@@ -16,10 +17,12 @@ impl<'a, T: 'a + Counter> Iter<'a, T> {
     pub fn new(hist: &'a Histogram<T>,
                valueUnitsPerBucket: u64)
                -> HistogramIterator<'a, T, Iter<'a, T>> {
+        assert!(valueUnitsPerBucket > 0);
         HistogramIterator::new(hist,
                                Iter {
                                    hist: hist,
                                    valueUnitsPerBucket: valueUnitsPerBucket,
+                                   // won't underflow because valueUnitsPerBucket > 0
                                    currentStepHighestValueReportingLevel: valueUnitsPerBucket - 1,
                                    currentStepLowestValueReportingLevel:
                                        hist.lowest_equivalent(valueUnitsPerBucket - 1),
--- a/src/iterators/linear.rs
+++ b/src/iterators/linear.rs
@@ -6,15 +6,15 @@ use iterators::{HistogramIterator, PickyIterator};
 pub struct Iter<'a, T: 'a + Counter> {
     hist: &'a Histogram<T>,
 
-    valueUnitsPerBucket: i64,
-    currentStepHighestValueReportingLevel: i64,
-    currentStepLowestValueReportingLevel: i64,
+    valueUnitsPerBucket: u64,
+    currentStepHighestValueReportingLevel: u64,
+    currentStepLowestValueReportingLevel: u64,
 }
 
 impl<'a, T: 'a + Counter> Iter<'a, T> {
     /// Construct a new linear iterator. See `Histogram::iter_linear` for details.
     pub fn new(hist: &'a Histogram<T>,
-               valueUnitsPerBucket: i64)
+               valueUnitsPerBucket: u64)
                -> HistogramIterator<'a, T, Iter<'a, T>> {
         HistogramIterator::new(hist,
                                Iter {
--- a/src/iterators/linear.rs
+++ b/src/iterators/linear.rs
@@ -17,7 +17,7 @@ impl<'a, T: 'a + Counter> Iter<'a, T> {
     pub fn new(hist: &'a Histogram<T>,
                valueUnitsPerBucket: u64)
                -> HistogramIterator<'a, T, Iter<'a, T>> {
-        assert!(valueUnitsPerBucket > 0);
+        assert!(valueUnitsPerBucket > 0, "valueUnitsPerBucket must be > 0");
         HistogramIterator::new(hist,
                                Iter {
                                    hist: hist,
--- a/src/iterators/log.rs
+++ b/src/iterators/log.rs
@@ -9,14 +9,14 @@ pub struct Iter<'a, T: 'a + Counter> {
     nextValueReportingLevel: f64,
     logBase: f64,
 
-    currentStepLowestValueReportingLevel: i64,
-    currentStepHighestValueReportingLevel: i64,
+    currentStepLowestValueReportingLevel: u64,
+    currentStepHighestValueReportingLevel: u64,
 }
 
 impl<'a, T: 'a + Counter> Iter<'a, T> {
     /// Construct a new logarithmic iterator. See `Histogram::iter_log` for details.
     pub fn new(hist: &'a Histogram<T>,
-               valueUnitsInFirstBucket: i64,
+               valueUnitsInFirstBucket: u64,
                logBase: f64)
                -> HistogramIterator<'a, T, Iter<'a, T>> {
         HistogramIterator::new(hist,
@@ -37,7 +37,7 @@ impl<'a, T: 'a + Counter> PickyIterator<T> for Iter<'a, T> {
         let val = self.hist.value_for(index);
         if val >= self.currentStepLowestValueReportingLevel || index == self.hist.last() {
             self.nextValueReportingLevel *= self.logBase;
-            self.currentStepHighestValueReportingLevel = self.nextValueReportingLevel as i64 - 1;
+            self.currentStepHighestValueReportingLevel = self.nextValueReportingLevel as u64 - 1;
             self.currentStepLowestValueReportingLevel = self.hist
                 .lowest_equivalent(self.currentStepHighestValueReportingLevel);
             true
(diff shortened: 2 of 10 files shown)


src/iterators/.percentiles.rs.rustfmt
```diff
@@ -1,89 +0,0 @@
-use num;
```
marshallpierce: this looks like a leftover rustfmt backup file

src/lib.rs
```diff
@@ -753,17 +749,21 @@ impl<T: Counter> Histogram<T> {
...
-    pub fn record_n_correct(&mut self, value: i64, count: T, interval: i64) -> Result<(), ()> {
+    pub fn record_n_correct(&mut self, value: u64, count: T, interval: u64) -> Result<(), ()> {
         try!(self.record_n(value, count));
         if interval <= 0 {
             return Ok(());
         }
 
-        let mut missingValue = value - interval;
-        while missingValue >= interval {
+        if value > interval {
```
marshallpierce: This was the only meaningful logic change.

tests/auto_resize.rs
```diff
@@ -9,19 +9,24 @@ use hdrsample::Histogram;
...
     let mut histogram = Histogram::<u64>::new(3).unwrap();
-    histogram += (1i64 << 62) - 1;
+    histogram += (1u64 << 62) - 1;
     assert_eq!(histogram.buckets(), 52);
     assert_eq!(histogram.len(), 54272);
-    histogram += i64::max_value();
-    assert_eq!(histogram.buckets(), 53);
-    assert_eq!(histogram.len(), 55296);
+    histogram += u64::max_value();
+    assert_eq!(histogram.buckets(), 54);
```
marshallpierce: This is the only meaningfully changed test

src/iterators/linear.rs
```diff
@@ -6,15 +6,15 @@ use iterators::{HistogramIterator, PickyIterator};
 pub struct Iter<'a, T: 'a + Counter> {
     hist: &'a Histogram<T>,
 
-    valueUnitsPerBucket: i64,
-    currentStepHighestValueReportingLevel: i64,
-    currentStepLowestValueReportingLevel: i64,
+    valueUnitsPerBucket: u64,
```
marshallpierce: Unrelated to this change, but there's a mix of camelCase and snake_case. Do you want me to make them all snake_case (as is the norm in rust-land) or leave them as is? I don't care strongly about it (I actually prefer camelCase visually, but I prefer to go with ecosystem norms even more than that), so if it's something you'd been meaning to get around to, I'll fix it, but otherwise I'll leave it.

src/iterators/log.rs
```diff
@@ -37,7 +37,7 @@ impl<'a, T: 'a + Counter> PickyIterator<T> for Iter<'a, T> {
         let val = self.hist.value_for(index);
         if val >= self.currentStepLowestValueReportingLevel || index == self.hist.last() {
             self.nextValueReportingLevel *= self.logBase;
-            self.currentStepHighestValueReportingLevel = self.nextValueReportingLevel as i64 - 1;
+            self.currentStepHighestValueReportingLevel = self.nextValueReportingLevel as u64 - 1;
```
jonhoo: Is this safe? Are we sure that `self.nextValueReportingLevel > 0`?

src/lib.rs
jonhoo: I wonder if this should actually be

```rust
let mut missingValue = value;
while missingValue >= interval {
    missingValue -= interval;
    try!(self.record_n(missingValue, count));
}
```

The semantics of that are slightly different (i.e., the "last" `interval` will also be counted), but I think that's how it is supposed to be (i.e., the old version was also wrong).

tests/auto_resize.rs
jonhoo: How did you arrive at 54?

jonhoo:
I'm happy to compare them to `master`. I wasn't planning on cutting another minor release any time soon, so preparing `master` for the eventual major release that this will be seems fine.

jonhoo:
My biggest worry about this change is the danger of `u64` underflow when we subtract. I've pointed out one already above, but there may be others. Could you make a pass and just check that there aren't other places where we're subtracting from a `u64` where we don't know that the result isn't negative?

tests/auto_resize.rs
marshallpierce: Max value is one power of 2 greater, so we need one more bucket.

src/iterators/log.rs
marshallpierce: Good catch, I'll double check this logic.

marshallpierce:
Sure, I'll check over it and try to convince myself that we won't underflow. My implementation used u64 and didn't have any issues, but you also have more functionality so there may well be things that don't play well.

tests/auto_resize.rs
jonhoo: oh, ofc, `u64::max_value() == i64::max_value() << 1`

jonhoo:
Yeah, I have this nagging sensation that there was some case where I couldn't figure out whether a particular number was allowed to be negative or not, and I just decided to do everything with `i64` (obviously a very lazy solution :p). I completely agree that the right thing to do (as you've started here) is to move to `u64` and figure out where, if anywhere, we need to watch for underflow.

marshallpierce:
I haven't forgotten about this; just been busy.

src/lib.rs
```diff
@@ -1250,7 +1250,7 @@ impl<T: Counter> Histogram<T> {
     /// of bins needed to cover that range.
     ///
     /// May fail if `high` is not at least twice the lowest discernible value.
-    fn cover(&mut self, high: i64) -> Result<usize, &'static str> {
+    fn cover(&mut self, high: u64) -> Result<usize, &'static str> {
         if high < 2 * self.lowestDiscernibleValue {
             return Err("highest trackable value cannot be < (2 * lowest discernible value)");
```
marshallpierce: This seems like programmer error, so I think this could probably be a `panic!` or `assert!`. Existing usage seems to suggest that this is the case. It's used in two places: new histograms, where this check cannot possibly fail as it's already validated, and `resize`, where the result is simply `unwrap()`'d.

src/iterators/linear.rs
```diff
@@ -6,20 +6,23 @@ use iterators::{HistogramIterator, PickyIterator};
...
+    currentStepLowestValueReportingLevel: u64,
 }
 
 impl<'a, T: 'a + Counter> Iter<'a, T> {
     /// Construct a new linear iterator. See `Histogram::iter_linear` for details.
     pub fn new(hist: &'a Histogram<T>,
-               valueUnitsPerBucket: i64)
+               valueUnitsPerBucket: u64)
                -> HistogramIterator<'a, T, Iter<'a, T>> {
+        assert!(valueUnitsPerBucket > 0);
```
marshallpierce: These, and similar ones in `log.rs`, can be `panic!`s if you think it's worth the extra couple of lines for hapless-end-user clarity.

src/lib.rs
```diff
@@ -753,17 +753,21 @@ impl<T: Counter> Histogram<T> {
...
-        let mut missingValue = value - interval;
-        while missingValue >= interval {
-            try!(self.record_n(missingValue, count));
-            missingValue -= interval;
+        if value > interval {
+            // only enter loop when calculations will stay non-negative
+            let mut missingValue = value - interval;
+            while missingValue >= interval {
+                try!(self.record_n(missingValue, count));
+                missingValue -= interval;
```
marshallpierce: Regarding your (now "outdated") comment on this loop -- I think not doing the last interval (in other words, the current logic) is correct, given the intent of this method.

Suppose the expected interval is 5, and the value we have is 17. The current logic would record 17, 12, 7, and stop: the lowest recorded value would be in `[expected interval, 2 * expected - 1]`. 

The alternate logic that includes an extra value would result in also recording the value 2. I think this is not right, because that's basically going to skew your latency numbers to be *better* than expected. It's certainly possible that in some cases this would be desirable, but I'd wager that in general users would be surprised by this (and not want it).

src/iterators/linear.rs
jonhoo: No, I'm happy to leave them as `assert!`. *Maybe* they should be `debug_assert!` instead though? What do you think?

src/lib.rs
jonhoo: Mmm, good point, I think you're right.

src/lib.rs
jonhoo: Yup, `assert!` here seems fine to me.

src/iterators/linear.rs
marshallpierce: Now that I think of it, no reason why I shouldn't provide an error message in the assert... Anyway, I think the performance impact will probably be minimal as it's only at the initialization of an iterator. The ones on `iterator/mod.rs:106` and the like are maybe of more concern, but even there I'm not too worried about it. I've found http://joeduffyblog.com/2016/02/07/the-error-model/#bugs-abandonment-assertions-and-contracts to be pretty persuasive about error handling in general, and it espouses a pro-assertion stance. 

I think for these sorts of assertions maybe `debug_assert!` is the configured the wrong direction by default. I want asserts on by default, even in release builds, and optionally removed for users who (1) know their code paths won't tickle the assertions and (2) need the last 2% of speed. We could have a `probably_assert!` (or whatever) that defaults to `assert!` unless a feature is set (`strip_asserts`?). I don't think we necessarily need this now, or maybe ever, but if benchmarks start showing assertions are painful, we have an easy out.

marshallpierce:
OK, I think that's everything -- let me know if there's anything else you'd like addressed in this PR.

jonhoo:
Looks good to me! Thanks for all the work you've put in :)

PR #3, opened 2016-11-29
Use u64 instead of T as the type for aggregated counts

marshallpierce:
While I work on the details of #1, here's a smaller fix I noticed.

With T as the total count, a Histogram<u8> will overflow on inserting the 256th value, even if no individual bucket overflows. This PR switches to using a u64 for the total count (and other associated local vars used in adding histograms, etc).


jonhoo:
Looks good -- thanks!

PR #7, opened 2017-02-13
Use snake case for var names. Enable a few rustc deny options.

marshallpierce:
I'll have some time to work on porting tests again soon so I figured I'd do this now when it's not likely to conflict. Mmm regexes.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1275,7 +1275,7 @@ impl<T: Counter> Histogram<T> {
         self.counts.resize(len, T::zero());
     }
 
-    /// Set internally tracked max value to new value if new value is greater than current one.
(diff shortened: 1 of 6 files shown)


src/lib.rs
```diff
@@ -137,8 +137,16 @@
...
+#![deny(
+    missing_docs,
+    trivial_casts,
+    trivial_numeric_casts,
+    unused_extern_crates,
+    unused_import_braces,
+    unused_results,
+    variant_size_differences,
+    warnings
+)]
```
marshallpierce: These seemed like nice things for rustc to check.

jonhoo:
lgtm!

PR #8, opened 2017-02-14
Tests for initialization logic

marshallpierce:
Tests for various configurations of min, max, and precision.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -206,6 +206,7 @@ impl<T> Counter for T
 /// = 2048 * 2^(k-1)`, which is the k-1'th bucket's end. So, we would use the previous bucket
 /// for those lower values as it has better precision.
 ///
+#[derive(Debug)]
 pub struct Histogram<T: Counter> {
     auto_resize: bool,
 
@@ -622,6 +623,13 @@ impl<T: Counter> Histogram<T> {
             1
         } - 1;
         let sub_bucket_count = 2_usize.pow(sub_bucket_half_count_magnitude as u32 + 1);
+
+        if unit_magnitude + sub_bucket_half_count_magnitude + 1 > 63 {
+            // Cannot represent shifted sub bucket count in 64 bits.
+            // This will cause an infinite loop when calculating number of buckets
+            return Err("Cannot represent significant figures' worth of measurements beyond lowest value");
+        };
+
         let sub_bucket_half_count = sub_bucket_count / 2;
         // sub_bucket_count is always at least 2, so subtraction won't underflow
         let sub_bucket_mask = (sub_bucket_count as u64 - 1) << unit_magnitude;
@@ -1431,3 +1439,6 @@ impl<T: Counter, F: Counter> PartialEq<Histogram<F>> for Histogram<T>
 // TODO: encoding/decoding
 // TODO: timestamps and tags
 // TODO: textual output
+
+#[cfg(test)]
+mod tests;
--- a/src/tests.rs
+++ b/src/tests.rs
@@ -1,0 +1,272 @@
+use super::Histogram;
+
+#[test]
+fn init_fields_smallest_possible_array() {
+    let h = histo64(1, 2, 0);
+
+    assert_eq!(2, h.highest_trackable_value);
+    assert_eq!(1, h.lowest_discernible_value);
+    assert_eq!(0, h.significant_value_digits);
+
+    assert_eq!(2, h.sub_bucket_count);
+    assert_eq!(1, h.sub_bucket_half_count);
+    assert_eq!(2, h.bucket_count);
+    // bottom full bucket, one more half bucket
+    assert_eq!(3, h.counts.len());
+    assert_eq!(0, h.sub_bucket_half_count_magnitude);
+    assert_eq!(1, h.sub_bucket_mask);
+
+    assert_eq!(0, h.unit_magnitude);
+    assert_eq!(0, h.unit_magnitude_mask);
+
+    assert_eq!(63, h.leading_zero_count_base);
+}
+
+#[test]
+fn init_fields_max_value_max_precision_largest_possible_array() {
+    let h = histo64(1, u64::max_value(), 5);
+
+    assert_eq!(u64::max_value(), h.highest_trackable_value);
+    assert_eq!(1, h.lowest_discernible_value);
+    assert_eq!(5, h.significant_value_digits);
+
+    // 5 sigdigs = 100,000. sub bucket must hold 200,000. 2^18 = 262,144.
+    assert_eq!(1 << 18, h.sub_bucket_count);
+    assert_eq!(1 << 17, h.sub_bucket_half_count);
+    // 2^46 * 2^18 = 2^64, so 47 buckets.
+    assert_eq!(47, h.bucket_count);
+    assert_eq!(46 * h.sub_bucket_half_count + h.sub_bucket_count, h.counts.len());
+    assert_eq!(17, h.sub_bucket_half_count_magnitude);
+    assert_eq!((1 << 18) - 1, h.sub_bucket_mask);
+
+    assert_eq!(0, h.unit_magnitude);
+    assert_eq!(0, h.unit_magnitude_mask);
+
+    assert_eq!(64 - 17 - 1, h.leading_zero_count_base);
+}
+
+#[test]
+fn init_fields_max_value_medium_precision() {
+    let h = histo64(1, u64::max_value(), 3);
+
+    assert_eq!(u64::max_value(), h.highest_trackable_value);
+    assert_eq!(1, h.lowest_discernible_value);
+    assert_eq!(3, h.significant_value_digits);
+
+    // should hit the case where it detects impending overflow
+    // 3 sigdigs = 1,000. sub bucket must hold 2,000. 2^11 = 2048.
+    assert_eq!(1 << 11, h.sub_bucket_count);
+    assert_eq!(1 << 10, h.sub_bucket_half_count);
+    // 2^53 * 2048 == 2^64, so that's 54 buckets.
+    assert_eq!(54, h.bucket_count);
+    assert_eq!(53 * h.sub_bucket_half_count + h.sub_bucket_count, h.counts.len());
+    assert_eq!(10, h.sub_bucket_half_count_magnitude);
+    assert_eq!((1 << 11) - 1, h.sub_bucket_mask);
+
+    assert_eq!(0, h.unit_magnitude);
+    assert_eq!(0, h.unit_magnitude_mask);
+
+    assert_eq!(64 - 10 - 1, h.leading_zero_count_base);
+}
+
+#[test]
+fn init_fields_1_bucket_medium_precision() {
+    let h = histo64(1, 2000, 3);
+
+    assert_eq!(2000, h.highest_trackable_value);
+    assert_eq!(1, h.lowest_discernible_value);
+    assert_eq!(3, h.significant_value_digits);
+
+    // 3 sigdigs = 1,000. sub bucket must hold 2,000. 2^11 = 2048.
+    assert_eq!(1 << 11, h.sub_bucket_count);
+    assert_eq!(1 << 10, h.sub_bucket_half_count);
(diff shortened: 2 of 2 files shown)


src/lib.rs
```diff
@@ -622,6 +623,13 @@ impl<T: Counter> Histogram<T> {
             1
         } - 1;
         let sub_bucket_count = 2_usize.pow(sub_bucket_half_count_magnitude as u32 + 1);
+
+        if unit_magnitude + sub_bucket_half_count_magnitude + 1 > 63 {
+            // Cannot represent shifted sub bucket count in 64 bits.
+            // This will cause an infinite loop when calculating number of buckets
+            return Err("Cannot represent significant figures' worth of measurements beyond lowest value");
```
marshallpierce: I'm pretty sure that the original Java is vulnerable to this issue as well, but probably nobody uses minimum values that are very very large.

tests/histogram.rs
```diff
@@ -1,7 +1,5 @@
 //! Tests from HistogramTest.java
 
-#![allow(non_snake_case)]
```
marshallpierce: Missed these ones in the prior cleanup.

src/lib.rs
```diff
@@ -206,6 +206,7 @@ impl<T> Counter for T
 /// = 2048 * 2^(k-1)`, which is the k-1'th bucket's end. So, we would use the previous bucket
 /// for those lower values as it has better precision.
 ///
+#[derive(Debug)]
```
jonhoo: Hmm, I wonder if we'd rather implement `Debug` ourselves. I feel like the autogenerated one won't be particularly helpful (or readable).

src/lib.rs
jonhoo: Good catch! Maybe report upstream?

tests/histogram.rs
jonhoo: I'd prefer these in a separate PR. I know it's a little annoying, but makes things cleaner.

jonhoo:
Just out of curiosity: what purpose are you envisioning that these added tests serve? Is it as a sanity check for if you refactor the code that computes offsets, magnitudes, etc.? They seem quite strict, so there isn't much room for things changing under the hood without these tests failing. Though I guess that's probably okay if we assume that the core histogram design is pretty much finalized at this point.

src/lib.rs
marshallpierce: Yeah, I will

tests/histogram.rs
marshallpierce: OK

src/lib.rs
marshallpierce: I only needed it to be able to `unwrap_err()` from `new`. I think it's probably good enough until someone complains that they want to have a better debug display.

marshallpierce:
> is it as a sanity check for if you refactor the code that computes offsets, magnitudes, etc.? 

Precisely. I should have made that more clear in the initial comments -- I'm about to follow up with a PR to use types that are as small as possible for fields like `bucket_count` that can be 8 bits instead of 64, etc. There are also one or two simplifications to the calculations that lead to those fields I want to try.

src/lib.rs
jonhoo: Ah, then I'd prefer those `unwrap_err`s to instead be `assert!(.is_err())` and remove this `derive`.

marshallpierce: I do want the actual error value, so I'd need a pretty clumsy `match` to get at it. What's the downside to deriving `Debug`? The default does the job, and it will let people use "{:?}" if they want to quickly println their histogram.

```
Histogram { auto_resize: false, highest_trackable_value: 10000, lowest_discernible_value: 1, significant_value_digits: 2, bucket_count: 7, sub_bucket_count: 256, sub_bucket_half_count: 128, sub_bucket_half_count_magnitude: 7, sub_bucket_mask: 255, leading_zero_count_base: 56, unit_magnitude: 0, unit_magnitude_mask: 0, max_value: 0, min_non_zero_value: 18446744073709551615, total_count: 0, counts: [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0] }
```

jonhoo: I did see the string matches, but I'd much prefer it if the library instead returned an `enum Error` where the strings we currently return in `Err` were instead tagged to the appropriate enum variants.

That all said, I guess I'd be okay with accepting a `derive(Debug)` and then have a later PR implement it more sanely. In particular, I think the output of `counts` is just wholly unhelpful. The same goes for having an `Error` enum; we could do that at a later time instead.

marshallpierce: Ah yes, an enum for the errors is on my to-do list. The string matching makes me sad, but I'd rather figure out the `new_*` refactoring first (and figure out what the error cases are).

jonhoo:
Can you rebase onto master (to get rid of 7cbf9e7), and then I'll merge?

marshallpierce:
Rebase is done btw.

jonhoo:
Looks good!

PR #12, opened 2017-02-15
Reduce memory footprint per histogram.

marshallpierce:
Size shrinks from 144 bytes (according to std::mem::size_of) to 112
on a 64-bit system.

I'm not done poring over these changes to increase my confidence that nothing is broken, but I figured I might as well get it in front of your eyes sooner rather than later.

I added some tests that write random values in a couple of different configurations because early on in these changes I hit some panics due to overflow, etc, and I wanted a little insurance that I'd solved them all. 

Also, adding `rand` for the random tests exposed some Cargo sadness. Out of the box, that wouldn't build for me because `num` and `rand` had a version clash (fixed with `cargo update`). Because Cargo.lock isn't version controlled, now we're in a situation where it builds for me and probably doesn't for you. I know crates.io suggests not version controlling Cargo.lock, but I think that's a mistake, and I wish they wouldn't suggest that, because now we have an inconsistent build. Not checking in the lockfile coupled with Cargo's default of sneakily not using exact matches for versions is a recipe for exactly this situation. :( Are you open to checking in Cargo.lock? Otherwise, for instance, we can't trust a CI build, let alone both build the same revision successfully / identically.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -297,7 +297,7 @@ impl<T: Counter> Histogram<T> {
     /// Find the bucket the given value should be placed in.
     ///
     /// May panic if the given value falls outside the current range of the histogram.
-    fn index_for(&self, value: u64) -> isize {
+    fn index_for(&self, value: u64) -> usize {
         let bucket_index = self.bucket_for(value);
         let sub_bucket_index = self.sub_bucket_for(value, bucket_index);
 
@@ -307,7 +307,7 @@ impl<T: Counter> Histogram<T> {
         // Calculate the index for the first entry that will be used in the bucket (halfway through
         // sub_bucket_count). For bucket_index 0, all sub_bucket_count entries may be used, but
         // bucket_base_index is still set in the middle.
-        let bucket_base_index = (bucket_index as isize + 1) << self.sub_bucket_half_count_magnitude;
+        let bucket_base_index = (bucket_index as usize + 1) << self.sub_bucket_half_count_magnitude;
 
         // Calculate the offset in the bucket. This subtraction will result in a positive value in
         // all buckets except the 0th bucket (since a value in that bucket may be less than half
@@ -315,19 +315,18 @@ impl<T: Counter> Histogram<T> {
         // twice as much space.
         let offset_in_bucket = sub_bucket_index as isize - self.sub_bucket_half_count as isize;
 
-        // The following is the equivalent of
-        // ((sub_bucket_index  - sub_bucket_half_count) + bucket_base_index;
-        (bucket_base_index + offset_in_bucket)
+        let index = bucket_base_index as isize + offset_in_bucket;
+        // This is always non-negative because offset_in_bucket is only negative (and only then by
+        // sub_bucket_half_count at most) for bucket 0, and bucket_base_index will be halfway into
+        // bucket 0's sub buckets in that case.
+        debug_assert!(index >= 0);
+        return index as usize;
     }
 
     /// Get a mutable reference to the count bucket for the given value, if it is in range.
     fn mut_at(&mut self, value: u64) -> Option<&mut T> {
         let i = self.index_for(value);
-        if i < 0 {
-            return None;
-        }
-
-        self.counts.get_mut(i as usize)
+        self.counts.get_mut(i)
     }
 
     // ********************************************************************************************
@@ -426,7 +425,7 @@ impl<T: Counter> Histogram<T> {
             // and add each non-zero value found at it's proper value:
 
             // Do max value first, to avoid max value updates on each iteration:
-            let other_max_index = source.index_for(source.max()) as usize;
+            let other_max_index = source.index_for(source.max());
             let other_count = source[other_max_index];
             self.record_n(source.value_for(other_max_index), other_count).unwrap();
 
@@ -618,7 +617,7 @@ impl<T: Counter> Histogram<T> {
         // largest value with single unit resolution, in [2, 200_000].
         let largest = 2 * 10_u32.pow(sigfig as u32);
 
-        let unit_magnitude = ((low as f64).log2() / 2_f64.log2()).floor() as u8;
+        let unit_magnitude = (low as f64).log2().floor() as u8;
         let unit_magnitude_mask = (1 << unit_magnitude) - 1;
 
         // We need to maintain power-of-two sub_bucket_count (for clean direct indexing) that is
@@ -628,15 +627,11 @@ impl<T: Counter> Histogram<T> {
         // that.
         // In [1, 18]. 2^18 > 2 * 10^5 (the largest possible
         // largest_value_with_single_unit_resolution)
-        let sub_bucket_count_magnitude = ((largest as f64).log2() / 2_f64.log2()).ceil() as u8;
-        let sub_bucket_half_count_magnitude = if sub_bucket_count_magnitude > 1 {
-            sub_bucket_count_magnitude
-        } else {
-            1
-        } - 1;
-        let sub_bucket_count = 2_u32.pow(sub_bucket_half_count_magnitude as u32 + 1);
+        let sub_bucket_count_magnitude = (largest as f64).log2().ceil() as u8;
+        let sub_bucket_half_count_magnitude = sub_bucket_count_magnitude - 1;
+        let sub_bucket_count = 1_u32 << (sub_bucket_count_magnitude as u32);
 
-        if unit_magnitude + sub_bucket_half_count_magnitude + 1 > 63 {
+        if unit_magnitude + sub_bucket_count_magnitude > 63 {
             // Cannot represent shifted sub bucket count in 64 bits.
             // This will cause an infinite loop when calculating number of buckets
             return Err("Cannot represent significant figures' worth of measurements beyond lowest value");
@@ -660,7 +655,7 @@ impl<T: Counter> Histogram<T> {
 
             // Establish leading_zero_count_base, used in bucket_index_of() fast path:
             // subtract the bits that would be used by the largest value in bucket 0.
-            leading_zero_count_base: (64 - unit_magnitude - sub_bucket_half_count_magnitude - 1),
+            leading_zero_count_base: 64 - unit_magnitude - sub_bucket_count_magnitude,
             sub_bucket_half_count_magnitude: sub_bucket_half_count_magnitude,
 
             unit_magnitude: unit_magnitude,
@@ -1089,7 +1084,7 @@ impl<T: Counter> Histogram<T> {
             return 100.0;
         }
 
-        let target_index = cmp::min(self.index_for(value), self.last() as isize) as usize;
+        let target_index = cmp::min(self.index_for(value), self.last());
         let total_to_current_index =
             (0..(target_index + 1)).map(|i| self[i]).fold(T::zero(), |t, v| t + v);
         100.0 * total_to_current_index.to_f64().unwrap() / self.total_count as f64
@@ -1107,8 +1102,8 @@ impl<T: Counter> Histogram<T> {
     /// May fail if the given values are out of bounds.
     pub fn count_between(&self, low: u64, high: u64) -> Result<T, ()> {
         use std::cmp;
-        let low_index = cmp::max(0, self.index_for(low)) as usize;
-        let high_index = cmp::min(self.index_for(high), self.last() as isize) as usize;
+        let low_index = self.index_for(low);
+        let high_index = cmp::min(self.index_for(high), self.last());
         Ok((low_index..(high_index + 1)).map(|i| self[i]).fold(T::zero(), |t, v| t + v))
     }
 
@@ -1121,7 +1116,7 @@ impl<T: Counter> Histogram<T> {
     /// May fail if the given value is out of bounds.
     pub fn count_at(&self, value: u64) -> Result<T, ()> {
         use std::cmp;
-        Ok(self[cmp::min(cmp::max(0, self.index_for(value)), self.last() as isize) as usize])
+        Ok(self[cmp::min(self.index_for(value), self.last())])
     }
 
     // ********************************************************************************************
@@ -1206,6 +1201,7 @@ impl<T: Counter> Histogram<T> {
         let bucket_index = self.bucket_for(value);
         let sub_bucket_index = self.sub_bucket_for(value, bucket_index);
         // calculate distance to next value
+        // TODO when is sub_bucket_index >= sub_bucket_count?
         1_u64 <<
         (self.unit_magnitude +
          if sub_bucket_index >= self.sub_bucket_count {
@@ -1219,27 +1215,27 @@ impl<T: Counter> Histogram<T> {
     // Internal helpers
     // ********************************************************************************************
 
-    /// Compute the lowest (and therefore highest precision) bucket index that can represent the
-    /// value.
+    /// Compute the lowest (and therefore highest precision) bucket index whose sub-buckets can
+    /// represent the value.
     #[inline]
     fn bucket_for(&self, value: u64) -> u8 {
         // Calculates the number of powers of two by which the value is greater than the biggest
         // value that fits in bucket 0. This is the bucket index since each successive bucket can
         // hold a value 2x greater. The mask maps small values to bucket 0.
-        (self.leading_zero_count_base - (value | self.sub_bucket_mask).leading_zeros() as u8)
+        self.leading_zero_count_base - (value | self.sub_bucket_mask).leading_zeros() as u8
     }
 
     #[inline]
-    /// Compute the position inside a bucket at which the given value should be recorded.
+    /// Compute the position inside a bucket at which the given value should be recorded, indexed
+    /// from position 0 in the bucket (in the first half, which is not used past the first bucket).
+    /// For bucket_index > 0, the result will be in the top half of the bucket.
     fn sub_bucket_for(&self, value: u64, bucket_index: u8) -> u32 {
-        // For bucket_index 0, this is just value, so it may be anywhere in 0 to sub_bucket_count. For
-        // other bucket_index, this will always end up in the top half of sub_bucket_count: assume
-        // that for some bucket k > 0, this calculation will yield a value in the bottom half of 0
-        // to sub_bucket_count. Then, because of how buckets overlap, it would have also been in the
-        // top half of bucket k-1, and therefore would have returned k-1 in bucket_index_of(). Since
-        // we would then shift it one fewer bits here, it would be twice as big, and therefore in
-        // the top half of sub_bucket_count.
-        // TODO: >>> ?
+        // Since bucket_index is simply how many powers of 2 greater value is than what will fit in
+        // bucket 0 (that is, what will fit in [0, sub_bucket_count)), we shift off that many
+        // powers of two, and end up with a number in [0, sub_bucket_count).
+        // For bucket_index 0, this is just value. For bucket index k > 0, we know value won't fit
+        // in bucket (k - 1) by definition, so this calculation won't end up in the lower half of
+        // [0, sub_bucket_count) because that would mean it would also fit in bucket (k - 1).
         (value >> (bucket_index + self.unit_magnitude)) as u32
     }
 
@@ -1286,6 +1282,7 @@ impl<T: Counter> Histogram<T> {
     ///
     /// May fail if `high` is not at least twice the lowest discernible value.
     fn cover(&mut self, high: u64) -> u32 {
+        // TODO validate before we get to here so we don't need an assert
         assert!(high >= 2 * self.lowest_discernible_value,
             "highest trackable value must be >= (2 * lowest discernible value)");
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1241,8 +1241,10 @@ impl<T: Counter> Histogram<T> {
 
     #[inline]
     fn value_from_loc(&self, bucket_index: u8, sub_bucket_index: u32) -> u64 {
-        // sum won't overflow; bucket_index and unit_magnitude are both <= 64.
-        // However, the resulting shift may overflow if unit magnitude is large, for instance.
+        // Sum won't overflow; bucket_index and unit_magnitude are both <= 64.
+        // However, the resulting shift may overflow given bogus input, e.g. if unit magnitude is
+        // large and the input sub_bucket_index is for an entry in the counts index that shouldn't
+        // be used (because this calculation will overflow).
         (sub_bucket_index as u64) << (bucket_index + self.unit_magnitude)
     }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1241,7 +1241,8 @@ impl<T: Counter> Histogram<T> {
 
     #[inline]
     fn value_from_loc(&self, bucket_index: u8, sub_bucket_index: u32) -> u64 {
-        // sum won't overflow; bucket_index and unit_magnitude are both <= 64
+        // sum won't overflow; bucket_index and unit_magnitude are both <= 64.
+        // However, the resulting shift may overflow if unit magnitude is large, for instance.
         (sub_bucket_index as u64) << (bucket_index + self.unit_magnitude)
     }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -215,23 +215,28 @@ pub struct Histogram<T: Counter> {
     // >= 1
     lowest_discernible_value: u64,
     // in [0, 5]
-    significant_value_digits: u32,
+    significant_value_digits: u8,
 
-    bucket_count: usize,
-    sub_bucket_count: usize,
-
-    leading_zero_count_base: u32,
-    sub_bucket_half_count_magnitude: usize,
-
-    unit_magnitude: usize,
-    sub_bucket_half_count: usize,
-
+    // in [1, 64]
+    bucket_count: u8,
+    // 2^(sub_bucket_half_count_magnitude + 1) = [2, 2^18]
+    sub_bucket_count: u32,
+    // sub_bucket_count / 2 = [1, 2^17]
+    sub_bucket_half_count: u32,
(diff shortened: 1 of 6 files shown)


src/lib.rs
```diff
@@ -657,17 +673,14 @@ impl<T: Counter> Histogram<T> {
...
+            // set by alloc() below
+            counts: Vec::new(),
         };
 
         // determine exponent range needed to support the trackable value with no overflow:
         let len = h.cover(high);
 
-        // Establish leading_zero_count_base, used in bucket_index_of() fast path:
-        // subtract the bits that would be used by the largest value in bucket 0.
-        h.leading_zero_count_base = (64 - h.unit_magnitude - h.sub_bucket_half_count_magnitude - 1) as u32;
```
jonhoo: I may be wrong here, but this used to be below for a reason. If memory serves me right, `cover` re-calculated either `unit_magnitude` or `sub_bucket_half_count_magnitude`, and so this had to be computed after that. But that may have changed now?

src/lib.rs
```diff
@@ -1117,14 +1130,27 @@ impl<T: Counter> Histogram<T> {
 
     /// Computes the matching histogram value for the given histogram bin.
     pub fn value_for(&self, index: usize) -> u64 {
+        // TODO what to do about indexes that are beyond what can be represented by this histogram?
+        // Pretty easy to end up shifting off the high end of a u64, e.g. asking for a big index
+        // when sigfigs is high.
```
jonhoo: Yeah, this is something I thought about back in the day, but didn't find a nice solution to. I guess the right thing to do is to make `value_for` return a `Result<u64>`, and then use `?` liberally in any users of `value_for`. I worry that this will make *everything* return `Result` though..

src/lib.rs
marshallpierce: Nope, it just does `bucket_count` and `highest_trackable_value` nowadays.

src/lib.rs
marshallpierce: That's not ideal from an ergonomics perspective, I agree... but I'm not sure we really have a choice if we want this to be usable in cases where panics are unacceptable, like in FFI-heavy code. I have just such a use-case in mind, and I'd be willing to do some `Result`-wrangling to avoid a panic.

tests/data_access.rs
```diff
@@ -28,7 +28,7 @@ struct Loaded {
 }
 
 const TRACKABLE_MAX: u64 = 3600 * 1000 * 1000;
-const SIGFIG: u32 = 3;
+const SIGFIG: u8 = 3;
```
jonhoo: Beyond the argument about what is a reasonable range for `SIGFIG`, is there a *technical* reason why we can't support more than 256 significant figures?

tests/histogram.rs
```diff
@@ -395,3 +398,50 @@ fn scaled_set_to() {
...
+
+    let mut rng = rand::weak_rng();
+
+    for _ in 0..1_000_000 {
+        let mut r: u64 = rng.gen();
+        if r == 0 {
+            r = 1;
+        }
+
+        h.record(r).unwrap();
```
jonhoo: You can use the shorthand `h += r` here. It will panic if `r` is out of range.

tests/histogram.rs
```diff
@@ -395,3 +398,50 @@ fn scaled_set_to() {
...
```
jonhoo: `h += r`

tests/histogram.rs
```diff
@@ -395,3 +398,50 @@ fn scaled_set_to() {
...
+#[test]
+fn random_write_middle_of_value_range_precision_3_no_panic() {
+    let low = 1_000;
+    let high = 1_000_000_000;
+    let mut h = Histogram::<u64>::new_with_bounds(low, high, 3).unwrap();
+
+    let mut rng = rand::weak_rng();
+
+    for _ in 0..1_000_000 {
+        h.record(rng.gen_range(low, high + 1)).unwrap();
```
jonhoo: `h += rng.gen_range(low, high + 1)`

tests/histogram.rs
```diff
@@ -395,3 +398,50 @@ fn scaled_set_to() {
     h2.set_to(&h1).unwrap();
     are_equal(&h1, &h2);
 }
+
+
+#[test]
+fn random_write_full_value_range_precision_5_no_panic() {
+    let mut h = Histogram::<u64>::new_with_bounds(1, u64::max_value(), 5).unwrap();
```
jonhoo: Why test specifically with `SIGFIG = 5`?

tests/histogram.rs
```diff
@@ -395,3 +398,50 @@ fn scaled_set_to() {
...
+
+#[test]
+fn random_write_middle_of_value_range_precision_3_no_panic() {
+    let low = 1_000;
+    let high = 1_000_000_000;
+    let mut h = Histogram::<u64>::new_with_bounds(low, high, 3).unwrap();
+
+    let mut rng = rand::weak_rng();
+
+    for _ in 0..1_000_000 {
```
jonhoo: It'd be better if this range was expressed in terms of `low` and `high`.

src/lib.rs
jonhoo: Ah, okay, great.

src/lib.rs
jonhoo: Maybe the way to go is to add auto-unwrapping shorthands (like `AddAssign`) that people can use if they are willing to have the code panic. I agree with you that it's better to have the *option* of getting a `Result` instead of a panic, so let's make that change.

jonhoo: And perhaps make that separately from this PR?

tests/data_access.rs
marshallpierce: Because it's capped at 5 in the data structure (e.g. largest single-unit-precision value is `2 * 10^5`). Technically it could go up to 18 before we overflow what we can do in a `u64`, but then it wouldn't be much of a `Hdr` histogram, just a regular old one. :)

tests/histogram.rs
marshallpierce: Just to hit the end of the precision range.

tests/histogram.rs
marshallpierce: ah, yup! Copy-pasta.

jonhoo:
I like this analysis. It's good to figure out exactly what precision is required for the various parameters, as it also gives us some additional guarantees from the compiler in terms of overflow/underflow detection. I don't know that the memory footprint change matters all that much, since `.counts` still dominates by far, but it's neat nonetheless.

As for checking in `Cargo.lock`, the crates.io FAQ has some interesting [observations](http://doc.crates.io/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries) about this. In particular, they note that one of the big reasons for not putting `Cargo.lock` in a library is that any *users* of the library now *also* need to take our specific version requirements into account. Which may mean double-including a shared dependency, or in the more extreme case, fail to compile due to conflicting version requirements. I'm honestly not sure what the right thing to do here is, but I'm leaning towards leaving `Cargo.lock` out for now since it's only required as a `dev-dependency`. AFAICT, Travis runs `cargo update`, and so should work just fine. Ideally I'd like to get rid of the dependency on `num` too at some point, but that's a whole other ordeal.

tests/data_access.rs
jonhoo: Oh, interesting... I wasn't aware of that at all. Do we check that `SIGFIG > 5` is never used somewhere currently?

tests/histogram.rs
marshallpierce: I think that syntax is unintuitive. Bias warning: I'm generally anti-operator-overloading because it lets everyone come up with their own mini-language, and that hasn't worked out well in C++. (You can end up so far in your own world of operator overloading that you end up having a hard time teaching a new team member how to write your custom variant of C++, etc.) No doubt because of that bias, it didn't even occur to me to use `+=`. To me, `+=` is a thing I can't trivially navigate to find the source of, and  unlike methods (in Rust anyway) it's also overloaded in the type sense: the same syntax can be used to add a histogram into the receiver of the method call, which makes it less comprehensible to my eyes.

jonhoo: Hehe, okay, fair enough.

tests/histogram.rs
marshallpierce: Oh wait, I take it back. It's not related to low and high; it's just "do something a million times". Is there a more idiomatic way to express that?

marshallpierce:
Yeah, that entry in the FAQ is very strange. It says:

> Users dependent on the library will not inspect the library’s Cargo.lock (even if it exists).

But also:

> If all libraries were to check in their Cargo.lock, then multiple copies of the library would be used, and perhaps even a version conflict.

Those can't both be true. Either `Cargo.lock` is used and can be a problem in the transitive dependency tree, or it isn't used, and it won't be a problem.

src/lib.rs
marshallpierce: Filed #13 to keep track of it.

tests/data_access.rs
marshallpierce: Yup: https://github.com/jonhoo/hdrsample/blob/master/src/lib.rs#L600

src/lib.rs
jonhoo: I wonder if we might want to do a quick PR fixing #13 first, as that would let you get rid of these `TODO`s for this PR.

tests/data_access.rs
jonhoo: It would be great to add a comment above that line with the explanation for that limitation.

tests/histogram.rs
jonhoo: Oh, you're right, I misread the code. It seems fine the way it stands. The `low` and `high` arguments to `gen_range` below is what really matters.

tests/data_access.rs
marshallpierce: I added a comment to the `SIGFIG`s in tests, but searching for that led me to the benchmarks, which I haven't been building. Criterion is, ah, "lightly documented". How should I run those bad boys?

src/lib.rs
marshallpierce: Once I'm done going over this (which might produce another `TODO` or two) I'll look into retroactively fixing those issues separately if it can be done tidily without replicating these type changes.

marshallpierce: It was getting complicated (for my brain, not the computer) to juggle multiple PRs so I picked a few small simplifications to include here since they fit well with the theme of closer type bounds.

src/lib.rs
```diff
@@ -292,7 +297,7 @@ impl<T: Counter> Histogram<T> {
     /// Find the bucket the given value should be placed in.
     ///
     /// May panic if the given value falls outside the current range of the histogram.
```
marshallpierce: Not removing this comment until I have more tests and better understanding of what happens when out-of-range values are used. Values that are too small will just get shifted until they're 0, so that's safe at least.

src/lib.rs
```diff
@@ -606,25 +614,24 @@ impl<T: Counter> Histogram<T> {
         // NOT ok to be +/- 2 units at 1999. Only starting at 2000. So internally, we need to
         // maintain single unit resolution to 2x 10^decimal_points.
 
-        // largest value with single unit resolution
-        let largest = 2 * 10_u64.pow(sigfig);
+        // largest value with single unit resolution, in [2, 200_000].
+        let largest = 2 * 10_u32.pow(sigfig as u32);
 
-        let unit_magnitude = ((low as f64).log2() / 2_f64.log2()).floor() as usize;
```
marshallpierce: `log_2(2)` is 1, so dividing by it isn't needed. The Java code used division with the natural log to calculate `log_2`, but fortunately we need not resort to such things.

src/lib.rs
```diff
@@ -606,25 +614,24 @@ impl<T: Counter> Histogram<T> {
...
+        let unit_magnitude = (low as f64).log2().floor() as u8;
         let unit_magnitude_mask = (1 << unit_magnitude) - 1;
 
         // We need to maintain power-of-two sub_bucket_count (for clean direct indexing) that is
         // large enough to provide unit resolution to at least
         // largest_value_with_single_unit_resolution. So figure out
-        // largest_value_with_single_unit_resolution's nearest power-of-two (rounded up), and use that:
-        let sub_bucket_count_magnitude = ((largest as f64).log2() / 2_f64.log2()).ceil() as usize;
```
marshallpierce: Same `log_2` simplification here.

src/lib.rs
```diff
@@ -606,25 +614,24 @@ impl<T: Counter> Histogram<T> {
...
-        let sub_bucket_half_count_magnitude = if sub_bucket_count_magnitude > 1 {
```
marshallpierce: `largest` is always at least 2, so `sub_bucket_count_magnitude` must be at least 1 (never 0). Therefore, the `else` branch is hit when `sub_bucket_count_magnitude` is 1, at which point it's the same as the other branch.

tests/data_access.rs
jonhoo: Hah, yes, Criterion unfortunately has some way to go in that department (and in some other aspects too). `benchmarks/busy.rs` is not really a benchmark per se, it just records lots and lots of values in a tight loop, which is useful for performance profiling. `benchmarks/all.rs` is the one that actually does profiling. It requires a little bit of juggling because `crates.io` does not allow `git` dependencies

```console
$ git diff Cargo.toml
diff --git a/Cargo.toml b/Cargo.toml
index e16fa3d..118d2f5 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,12 +20,12 @@ license = "MIT/Apache-2.0"
 travis-ci = { repository = "jonhoo/hdrsample" }

 [features]
-#benchmark = ["criterion"]
-benchmark = [] # for crates.io publication
+benchmark = ["criterion"]
+#benchmark = [] # for crates.io publication

 [dependencies]
 num = "0.1"
-#criterion = { git = "https://github.com/japaric/criterion.rs.git", optional = true }
+criterion = { git = "https://github.com/japaric/criterion.rs.git", optional = true }

 [lib]
 path = "src/lib.rs"
$ rustup run nightly cargo run --bin perf --release --features benchmark
```

Criterion also does performance regression testing, so it might report a "failure" in some cases. Just ignore that. What is interesting is the performance numbers it spits out. On my machine, I get

```
Benchmarking raw_record_with_interval_u64
> Warming up for 2.0000 s
> Collecting 7000 samples in estimated 5.0201 s
> Found 211 outliers among 6999 measurements (3.01%)
  > 117 (1.67%) high mild
  > 94 (1.34%) high severe
> Performing linear regression
  >  slope [11.496 ns 11.514 ns]
  >    R^2  0.1003064 0.1002968
> Estimating the statistics of the sample
  >   mean [11.566 ns 11.668 ns]
  > median [11.590 ns 11.590 ns]
  >    MAD [12.385 ps 17.341 ps]
  >     SD [558.04 ps 3.9870 ns]

Benchmarking raw_record_u64
> Warming up for 1.0000 s
> Collecting 7000 samples in estimated 5.1764 s
> Found 263 outliers among 6999 measurements (3.76%)
  > 97 (1.39%) high mild
  > 166 (2.37%) high severe
> Performing linear regression
  >  slope [11.724 ns 11.763 ns]
  >    R^2  0.0296366 0.0296345
> Estimating the statistics of the sample
  >   mean [11.738 ns 11.776 ns]
  > median [11.645 ns 11.678 ns]
  >    MAD [406.89 ps 445.48 ps]
  >     SD [746.95 ps 880.17 ps]
```

src/lib.rs
```diff
@@ -641,10 +648,14 @@ impl<T: Counter> Histogram<T> {
...
-            bucket_count: 0, // set by establish_size below
+            // set by cover() below
+            bucket_count: 0,
             sub_bucket_count: sub_bucket_count,
 
-            leading_zero_count_base: 0, // set below, needs establish_size
+
+            // Establish leading_zero_count_base, used in bucket_index_of() fast path:
+            // subtract the bits that would be used by the largest value in bucket 0.
+            leading_zero_count_base: 64 - unit_magnitude - sub_bucket_count_magnitude,
```
marshallpierce: Now  that `sub_bucket_count_magnitude` is simply ``sub_bucket_half_count_magnitude + 1`, we can use it here.

src/lib.rs
```diff
@@ -1094,8 +1102,8 @@ impl<T: Counter> Histogram<T> {
     /// May fail if the given values are out of bounds.
     pub fn count_between(&self, low: u64, high: u64) -> Result<T, ()> {
         use std::cmp;
-        let low_index = cmp::max(0, self.index_for(low)) as usize;
```
marshallpierce: `max(0, x_u64) = x`

src/lib.rs
```diff
@@ -1180,6 +1201,7 @@ impl<T: Counter> Histogram<T> {
         let bucket_index = self.bucket_for(value);
         let sub_bucket_index = self.sub_bucket_for(value, bucket_index);
         // calculate distance to next value
+        // TODO when is sub_bucket_index >= sub_bucket_count?
```
marshallpierce: Any ideas?

src/lib.rs
```diff
@@ -1193,37 +1215,38 @@ impl<T: Counter> Histogram<T> {
...
+    /// Compute the position inside a bucket at which the given value should be recorded, indexed
+    /// from position 0 in the bucket (in the first half, which is not used past the first bucket).
+    /// For bucket_index > 0, the result will be in the top half of the bucket.
+    fn sub_bucket_for(&self, value: u64, bucket_index: u8) -> u32 {
+        // Since bucket_index is simply how many powers of 2 greater value is than what will fit in
+        // bucket 0 (that is, what will fit in [0, sub_bucket_count)), we shift off that many
+        // powers of two, and end up with a number in [0, sub_bucket_count).
+        // For bucket_index 0, this is just value. For bucket index k > 0, we know value won't fit
+        // in bucket (k - 1) by definition, so this calculation won't end up in the lower half of
+        // [0, sub_bucket_count) because that would mean it would also fit in bucket (k - 1).
```
marshallpierce: Reworded to be a little more concrete.

jonhoo:
@marshallpierce this is starting to look pretty good! :+1:

src/lib.rs
```diff
@@ -1249,15 +1272,17 @@ impl<T: Counter> Histogram<T> {
     /// covered by previous buckets), and the +1 being used for the lower half of the 0'th bucket.
     /// Or, equivalently, we need 1 more bucket to capture the max value if we consider the
     /// sub-bucket length to be halved.
-    fn num_bins(&self, number_of_buckets: usize) -> usize {
-        (number_of_buckets + 1) * (self.sub_bucket_count / 2)
+    fn num_bins(&self, number_of_buckets: u8) -> u32 {
+        // TODO use sub_bucket_half_count
+        (number_of_buckets as u32 + 1) * (self.sub_bucket_count / 2)
```
marshallpierce: There might be a subtle issue with this being a "cold" field, so I'll see if this gets accepted as an upstream PR before I do it here. (Also, would be interesting to see if we can get a win by controlling field order once we get to concurrent histograms.)

marshallpierce:
OK, I think that's as much as I want to squeeze into this PR. Double-check my reasoning, if you don't mind! :)

src/lib.rs
```diff
@@ -606,25 +614,24 @@ impl<T: Counter> Histogram<T> {
...
-        let sub_bucket_count = 2_usize.pow(sub_bucket_half_count_magnitude as u32 + 1);
-
-        if unit_magnitude + sub_bucket_half_count_magnitude + 1 > 63 {
+        // largest_value_with_single_unit_resolution's nearest power-of-two (rounded up), and use
+        // that.
+        // In [1, 18]. 2^18 > 2 * 10^5 (the largest possible
+        // largest_value_with_single_unit_resolution)
+        let sub_bucket_count_magnitude = (largest as f64).log2().ceil() as u8;
+        let sub_bucket_half_count_magnitude = sub_bucket_count_magnitude - 1;
+        let sub_bucket_count = 1_u32 << (sub_bucket_count_magnitude as u32);
```
marshallpierce: More half_count + 1 => count. Also, since `2^k == 1 << k`, this seems more consistent with all the other bitshifting going on.

src/lib.rs
jonhoo: Yeah, I reasoned my way to that when I saw the change.

src/lib.rs
jonhoo: Hmm, is that right?
```
let sub_bucket_half_count_magnitude = sub_bucket_count_magnitude - 1;
h.leading_zero_count_base = (64 - h.unit_magnitude - h.sub_bucket_half_count_magnitude - 1) as u32;
leading_zero_count_base = (64 - unit_magnitude - sub_bucket_count_magnitude - 1 - 1) as u32;
leading_zero_count_base = (64 - unit_magnitude - sub_bucket_count_magnitude - 2) as u32;
```

src/lib.rs
```diff
@@ -1193,37 +1215,38 @@ impl<T: Counter> Histogram<T> {
...
+        // For bucket_index 0, this is just value. For bucket index k > 0, we know value won't fit
+        // in bucket (k - 1) by definition, so this calculation won't end up in the lower half of
+        // [0, sub_bucket_count) because that would mean it would also fit in bucket (k - 1).
+        (value >> (bucket_index + self.unit_magnitude)) as u32
     }
 
     #[inline]
-    fn value_from_loc(&self, bucket_index: usize, sub_bucket_index: usize) -> u64 {
+    fn value_from_loc(&self, bucket_index: u8, sub_bucket_index: u32) -> u64 {
+        // sum won't overflow; bucket_index and unit_magnitude are both <= 64
```
jonhoo: Could still overflow depending on the value of `sub_bucket_index` though, right?

marshallpierce: Yes, though actually because of `unit_magnitude`. The sum of the two `u8`'s won't overflow, but because `unit_magnitude` might be, say, 62, then the `sub_bucket_index` (which is at most 18 bits) might be entirely shifted off the high end of a `u64`.

We should carefully audit code paths to make sure that they won't hit this when addressing #13, but at least it's no more fragile than it was before.

src/lib.rs
jonhoo: With `sub_bucket_half_count` being cold you mean? Yeah, that's probably right, though it's unclear just how much performance would be affected by it.. Why is it you would like to use `sub_bucket_half_count` instead?

src/lib.rs
jonhoo: I'm actually not sure when this would be triggered. If I remember correctly, this is also present in the HdrHistogram code, and I just ported if over directly for fear of breaking something...

src/lib.rs
marshallpierce: To save an integer division, which http://www.agner.org/optimize/#manuals tells me has nontrivial latency.

marshallpierce: We'll see what Mr HdrHistogram himself thinks. https://github.com/HdrHistogram/HdrHistogram/pull/119

src/lib.rs
jonhoo: Sort of unrelated to this particular change (or PR), but I'd like at some point for us to write up in a more understandable way how the end-to-end histogram works. Like, explain the whole notion of buckets, half buckets, subcounts, etc. Would be useful both for future contributors, and to keep ourselves sane :p

src/lib.rs
jonhoo: One alternative here might be to use `>> 1` instead, though I don't know that that'll actually end up being faster.

src/lib.rs
marshallpierce: Yes! I've been thinking that at the end of this revisiting of the logic, it would be great to write up basically a design document. There was some talk of having it in a centralized place along with some test vectors in a dedicated repo or something, so I know Gil is open to the idea.

In particular I think it would be interesting to have both the slow version and optimized version of the core routines because it's often hard to reverse engineer the high level intent from all the bit wrangling.

src/lib.rs
jonhoo: Yeah, that's the overflow I was referring to.

I agree that we should audit the code path, but I feel like the comment for now is inaccurate, given that it's not clear that it's true it won't overflow.

src/lib.rs
marshallpierce: True. The compiler may well be smart enough to map a integer division by 2 to that anyway, actually... but worth a try if Gil points out some sneaky reason why we want this instead of the field that already has the right answer in it.

src/lib.rs
marshallpierce: No, the comment is quite accurate -- a little too sneaky, perhaps: the *sum* won't overflow, but the shift may. I'll add a note about that. :)

jonhoo: Haha, fair enough... Clarifying that would be good.

src/lib.rs
```diff
@@ -1241,7 +1241,8 @@ impl<T: Counter> Histogram<T> {
 
     #[inline]
     fn value_from_loc(&self, bucket_index: u8, sub_bucket_index: u32) -> u64 {
-        // sum won't overflow; bucket_index and unit_magnitude are both <= 64
+        // sum won't overflow; bucket_index and unit_magnitude are both <= 64.
+        // However, the resulting shift may overflow if unit magnitude is large, for instance.
```
jonhoo: nit, but either capitalize "sum" above, or downcase "However".

marshallpierce: Good point, and I took the opportunity to make it less scary.

jonhoo:
"bogus" input is a bit strong, but I'll take it. LGTM. I'm happy to merge if you feel finished with it.

marshallpierce:
Yep, I think it's done. On to the next! 

I felt "bogus" was warranted because you could, for instance, configure a histogram with a lowest value of 1 << 62 and a precision of 5. This will actually only be able to meaningfully use *2* slots before it overflows `u64`, but if you ask `value_for` (which is public) for the value for index 100, which I would say is bogus, you will get an equally bogus answer. Good old GIGO...

Issue #10, opened 2017-02-14
Use an enum rather than strings for errors


marshallpierce:
This will make it nicer for testing, and more comprehensible for users that may want to handle individual failure cases.

jonhoo:
Fixed in 62575317531b14d01e20f028b98de4c687e3d1df following #22.

PR #22, opened 2017-02-20
Error enums

marshallpierce:


marshallpierce:
This is on top of #16.

src/lib.rs
```diff
@@ -242,6 +242,42 @@ pub struct Histogram<T: Counter> {
...
+    /// Lowest discernible value must be >= 1
+    LowTooSmall,
+    /// Lowest discernible value must be <= u64::max_value() / 2
+    LowTooBig,
+    /// Highest trackable value must be >= 2 * lowest discernible value
+    HighLessThanTwiceLow,
+    /// Number of significant digits must be between 0 and 5
+    SigFigTooBig,
+    /// Cannot represent sigfig worth of values beyond low
+    CannotRepresentSigFigBeyondLow
```
marshallpierce: I expect most people will probably not match each of these variants but at least it will be nicely specific if their usage panics.

src/lib.rs
```diff
@@ -242,6 +242,42 @@ pub struct Histogram<T: Counter> {
...
+}
+
+/// Errors that can occur when adding another histogram.
+#[derive(Debug, Eq, PartialEq, Clone, Copy)]
+pub enum AdditionError {
+    /// The other histogram includes values that do not fit in this histogram's range.
+    /// Only possible when auto resize is disabled.
+    OtherAddendValuesExceedRange
```
marshallpierce: Annoyingly, "addend" refers to both terms in an addition. The first one is called the "augend" but that's not what we care about. Oddly this value thing means that addition of non-resizing histograms is not commutative.

src/lib.rs
```diff
@@ -242,6 +242,42 @@ pub struct Histogram<T: Counter> {
...
+/// Errors that can occur when subtracting another histogram.
+#[derive(Debug, Eq, PartialEq, Clone, Copy)]
+pub enum SubtractionError {
+    /// The other histogram includes values that do not fit in this histogram's range.
+    /// Only possible when auto resize is disabled.
+    SubtrahendValuesExceedMinuendRange,
+    /// The other histogram includes counts that are higher than the current count for a value, and
+    /// counts cannot go negative. The subtraction may have been partially applied to some counts as
+    /// this error is returned when the first impossible subtraction is detected.
+    SubtrahendCountExceedsMinuendCount
```
marshallpierce: For subtraction we can have maximally precise (or pedantic) names, though!

marshallpierce:
This addresses #10.

src/lib.rs
jonhoo: Yup. I think we might be able to bikeshed some better names for the variants, but it's probably not worth the time.

I do think we should give some better documentation for the variants though. This will be one of the *main* places where people will look for an explanation of why something didn't work, so we should make sure things are as clear as possible. For example, *why* can't SigFig be bigger than 5, *why* must high be at least twice of low, etc. For some of these we have explanations elsewhere in the code, and I think those can be moved here.

Also, a minor point, but you should enclose code things (like `u64::max_value() / 2`) in backticks.

src/lib.rs
jonhoo: Huh, interesting. I think that's probably fine though. If I remember correctly, this is the reason I implement only `AddAssign`, and not `Add`.

jonhoo:
Looking good! I'd like some better docs for the `Creation` errors, because I suspect people will run into them more often than we think, and that's where they'll look to get an explanation, but beyond that I'm happy.

marshallpierce:
I took the liberty of self-bikeshedding on enum names, and also centralized the docs for the `new_*` methods to make it easier to maintain details on `sigfigs` and things. The internal documentation for the "sig figs + low = too many bits" error referred to things that aren't exposed externally (like `unit_magnitude`) so I opted not to just move that up to the corresponding variant.

jonhoo:
Looks good! Fixes #10.

Issue #19, opened 2017-02-20
Criterion data in repo


marshallpierce:
Is it right that `.criterion` is checked in? It seems like it's using that as a baseline to compare against or something, which isn't applicable on my hardware. Also, presumably if you were benchmarking against a baseline, wouldn't you want to set the baseline on a particular revision, then apply the changes under consideration?

jonhoo:
Yeah, Criterion is *supposed* to do regression testing whenever you run it, and it uses `.criterion` for this. And they advice that you should check it in. However, given how broken it currently is, plus the fact that it doesn't take into account running on different machines, I think we can just delete it.

jonhoo:
Closed by #23.

PR #23, opened 2017-02-20
Remove criterion data as per #19

marshallpierce:
That's a whole lot of data. If only there were some way to record data compactly with configurable accuracy...


jonhoo:
By the way, if you put "Fixes #<PR number>" in the commit message, GitHub will automatically close the corresponding issue when the PR is merged.

marshallpierce:
Ah yeah, I couldn't remember what that syntax was, and then got distracted and forgot to look it up before making that commit. Doh.

PR #27, opened 2017-03-02
Allow optional serialization of Histogram

andrewjstone:
Serialization can be derived via either serde or rustc-serialize
depending upon feature flags.

To enable serialization via serde:
cli: `cargo build --features "serde-serialize"`
Cargo.toml: `features = ["rustc-serialize"]`

To enable serialization via rustc-serialize:
cli: `cargo build --features "rustc-serialize"`
Cargo.toml: `features = ["serde-serialize"]`

Also remove TODO for serialization/encoding from README.md

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,9 +22,13 @@ travis-ci = { repository = "jonhoo/hdrsample" }
 [features]
 #benchmark = ["criterion"]
 benchmark = [] # for crates.io publication
+serde-serialize = ["serde", "serde_derive"]
 
 [dependencies]
 num = "0.1"
+rustc-serialize = { optional = true, version = "0.3" }
+serde = { optional = true, version = "0.9" }
+serde_derive = { optional = true, version = "0.9" }
 #criterion = { git = "https://github.com/japaric/criterion.rs.git", optional = true }
 
 [dev-dependencies]
--- a/README.md
+++ b/README.md
@@ -130,7 +130,6 @@ not (yet) been implemented:
    `DoubleHistogram` features are supported.
  - The `Recorder` feature of HdrHistogram.
  - Value shifting ("normalization").
- - Histogram serialization and encoding/decoding.
  - Timestamps and tags.
  - Textual output methods. These seem almost orthogonal to HdrSample, though it might be
    convenient if we implemented some relevant traits (CSV, JSON, and possibly simple
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -150,6 +150,13 @@
 
 extern crate num;
 
+#[cfg(feature = "rustc-serialize")]
+extern crate rustc_serialize;
+
+#[cfg(feature = "serde-serialize")]
+#[macro_use]
+extern crate serde_derive;
+
 use std::borrow::Borrow;
 use std::cmp;
 use std::ops::{Index, IndexMut, AddAssign, SubAssign};
@@ -204,6 +211,8 @@ impl<T> Counter for T
 /// = 2048 * 2^(k-1)`, which is the k-1'th bucket's end. So, we would use the previous bucket
 /// for those lower values as it has better precision.
 ///
+#[cfg_attr(feature = "rustc-serialize",  derive(RustcEncodable, RustcDecodable))]
+#[cfg_attr(feature = "serde-serialize",  derive(Serialize, Deserialize))]
 #[derive(Debug)]
 pub struct Histogram<T: Counter> {
     auto_resize: bool,


jonhoo:
Thanks for the PR!

While we could `derive(Serialize, Deserialize)`, we want to conform to the HdrHistogram [serialization format](https://hdrhistogram.github.io/HdrHistogram/JavaDoc/org/HdrHistogram/HistogramLogReader.html), both because it is way more space-efficient (it avoids storing the 0 counts in the trailing buckets), and because it gives us interoperability with other HdrHistogram tools/implementations. @marshallpierce, thoughts?

marshallpierce:
I think it would create user confusion to have multiple plausible-looking ways of serializing, so I'd say let's hold off on this. I hope to get to the sorta-zig-zag encoded form in the next week or two, so it shouldn't be long to wait.

jonhoo:
I still think we should expose serialization using the `Serialize` and `Deserialize` traits if at all possible though. That would let our histograms be sent much easier in various serialization contexts.

marshallpierce:
OK. As long as we can carefully document it so it's clear to users, I'm fine with enabling better interop with users wanting to serialize this as part of a larger struct, etc.

andrewjstone:
Hmm. That makes some sense, but I also think they are somewhat orthogonal concerns. For sending over the network in a predefined format (msgpack, protobuf, etc..) a Histogram, or a different structure that represents the log format in your link still has to implement `Serialize/Deserialize` so that it can use standard rust libraries to encode the structures. 

While it may be more efficient to construct the log format and then implement `Serialize/Deserialize` on that, it is by no means a given. Especially since compression of the serialized data would certainly remove the extra zeroes. 

That being said, there is still much value in adhering to the log format for interoperability with other tools, but as far as I can see that isn't really the same as the abstract serialization provided here. For instance, what is the "compressed HdrHistogram" defined here: "Interval_Compressed_Histogram: The fourth field must contain a text field parse-able as a Base64 text representation of a compressed HdrHistogram."

andrewjstone:
Seems like we overlapped, and all have the same goals of doing the right thing and allowing:

1) serialization to standardize log format
2) implementing Serialize/Deserialize

They could definitely be combined, and I wouldn't even mind implementing the log format and then implementing serialize/deserialize on that. However, that page doesn't seem to be quite descriptive enough for me. It's just all text fields. So I guess we'd really be just serializing/deserializing a string of those fields, which seems odd. And do you want to then convert the deserialized value into the histogram again?

marshallpierce:
There is a custom compressed binary format implemented by the Java version (and others). The interval log uses base64 of that binary format to express things like one histogram per minute so you can see trends in performance over time.

The serialization of interest here is the custom binary format. I have some progress towards implementing it sitting in another Rust implementation that I started on a while ago and then decided to merge with this one. 

Producing the "interval log" format, or consuming it, would be the responsibility of a benchmarking tool or analysis tool respectively, not so much a concern of this data structure.

andrewjstone:
Ahh Ok. I think I got it. So really you want to just serialize to the binary format so it can be read by tools. That format may be sent over the wire as a blob but you don't ever expect to deserialize it back into a Histogram. 

Unfortunately I need to be able to deserialize back into the Histogram struct. If that blob can be deserialized back into a histogram that works for me. If not, it seems like you could just add a function like log_serialize() that generates the Vec<u8> and then also have the Serialize/Deserialiize as in this PR.

My usecase is implementing status in an actor system across a cluster of nodes. I want to be able to gather histograms on a single node that has a connected client and then run user defined calculations (such as get X percentile) for each node's histogram and then return that to the user. While it's possible to do such a thing by just forwarding the user request to all nodes, running the calculation and returning it, which may indeed be more efficient, I'd like to make the histograms serializable, so that, like all other messages it can be sent remotely and locally.

This being merged isn't super critical to me right now, so of course I'd like you guys who seem to have much more experience with hdrhistogram to do it the right way. Thanks for taking the time to discuss this with me. I gotta run out for a few hours now.

marshallpierce:
The custom binary serialization is exactly what you want. :) It is designed for precisely that use case: efficient shoving-over-the-wire.

Since you actually have a need for it, I'll pause my investigation into normalization (another feature this implementation doesn't yet have) and work on serialization instead.

marshallpierce:
To be clear -- deserialization will also be supported (which will yield a histogram identical to the source). The binary format is only meant for efficient transport, so the receiver would presumably deserialize it to perform further analysis, which sounds like what you're intending to do.

andrewjstone:
Awesome @marshallpierce! Thanks so much. I'm glad I did the little work to get this PR in, cause I learned a bunch :) Closing in favor of your solution.

jonhoo:
For posterity: fixed in #28.

PR #28, opened 2017-03-09
Serialization

marshallpierce:
I think this is now probably in a usable state.

Things I still want to do (not necessarily in this PR):
- Optimization, especially around deserialization. Serialization is about the same speed as the Java impl, but deserialization is half as fast as Java.
- Interop testing. There is no corpus of test data available (yet), so I'll need to go make that first, but I would like to have an organized way of making sure this impl behaves identically to the Java one.
- Compressed format. This only implements V2, not V2Compressed.


src/serialization/serialization.rs
```diff
@@ -0,0 +1,409 @@
...
+        self.buf.write_u64::<BigEndian>(h.highest_trackable_value)?;
+        // int to double conversion
+        self.buf.write_f64::<BigEndian>(1.0)?;
+
+        debug_assert_eq!(V2_HEADER_SIZE, self.buf.len());
+
+        unsafe {
+            // want to treat the rest of the vec as a slice, and we've already reserved this
+            // space, so this way we don't have to resize() on a lot of dummy bytes.
+            self.buf.set_len(max_size);
```
marshallpierce: There area few uses of unsafe, so eye them with due suspicion -- this one is a 40% serialization speed boost.

src/serialization/serialization.rs
```diff
@@ -0,0 +1,409 @@
...
+        }
+
+        let counts_len = encode_counts(h, &mut self.buf[V2_HEADER_SIZE..])?;
+        let total_len = V2_HEADER_SIZE + counts_len;
+
+        // TODO benchmark fastest buffer management scheme
```
marshallpierce: I'm not yet ready to give up on serialization performance, so this is just a note to future us.

Cargo.toml
```diff
@@ -22,9 +22,11 @@ travis-ci = { repository = "jonhoo/hdrsample" }
 [features]
 #benchmark = ["criterion"]
 benchmark = [] # for crates.io publication
+bench_private = [] # for enabling nightly-only feature(test) on the main crate to allow benchmarking private code
```
jonhoo: Any particular reason why you don't reuse `benchmark` for this?

src/lib.rs
```diff
@@ -1503,3 +1516,7 @@ impl<T: Counter, F: Counter> PartialEq<Histogram<F>> for Histogram<T>
 #[path = "tests/tests.rs"]
 #[cfg(test)]
 mod tests;
+
+/// Serialization/deserialization support.
+#[path = "serialization/serialization.rs"]
```
jonhoo: Any particular reason you don't name the file `serialization/mod.rs` as is normal for Rust modules, which will let you get rid of this `#[path]` directive?

src/serialization/serialization.rs
```diff
@@ -0,0 +1,412 @@
...
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+pub enum SerializeError {
+    /// A count above i64::max_value() cannot be zig-zag encoded, and therefore cannot be
+    /// serialized.
+    CountNotSerializable,
+    /// Internal calculations cannot be represented in `usize`. Use smaller histograms or beefier
+    /// hardware.
+    UsizeTypeTooSmall,
+    /// An i/o operation failed.
+    IoError
```
jonhoo: Maybe have this wrap the `std::io::Error` instead of just throwing it away?

src/serialization/serialization.rs
```diff
@@ -0,0 +1,412 @@
...
+}
+
+impl std::convert::From<std::io::Error> for SerializeError {
+    fn from(_: std::io::Error) -> Self {
+        SerializeError::IoError
+    }
+}
+
+/// Serializer for the V2 binary format.
+/// Serializers are intended to be re-used for many histograms.
```
jonhoo: What is the reasoning behind having a separate `struct` for this, rather than just having methods on `Histogram`? You could still keep the implementation in a separate file by adding a trait (i.e., `Serialize`), and then implementing it for `Histogram`.

src/serialization/serialization.rs
```diff
@@ -0,0 +1,412 @@
...
+        // therefore be about 5e7 (50 million) at most.
+        length.checked_mul(9)
+    }
+}
+
+/// Errors that can happen during deserialization.
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+pub enum DeserializeError {
+    /// An i/o operation failed.
+    IoError,
```
jonhoo: Keep the underlying `std::io::Error`

src/lib.rs
```diff
@@ -162,12 +165,12 @@ use iterators::HistogramIterator;
 /// into an integer count. Partial ordering is used for threshholding, also usually in the context
 /// of percentiles.
 pub trait Counter
-    : num::Num + num::ToPrimitive + num::FromPrimitive + Copy + PartialOrd<Self> {
+    : num::Num + num::ToPrimitive + num::FromPrimitive + num::Bounded + Copy + PartialOrd<Self> {
```
jonhoo: Why is this new bound needed?

src/serialization/serialization.rs
```diff
@@ -0,0 +1,412 @@
...
+        let int_double_ratio = reader.read_f64::<BigEndian>()?;
+        if int_double_ratio != 1.0 {
+            return Err(DeserializeError::UnsupportedFeature);
+        }
+
+        let mut h = Histogram::new_with_bounds(low, high, num_digits)
+            .map_err(|_| DeserializeError::InvalidParameters)?;
+
+        if payload_len > self.payload_buf.len() {
+            self.payload_buf.resize(payload_len, 0);
```
jonhoo: Shouldn't the `payload_buf` always be resized? Otherwise it might be *longer* than the underlying data.

src/serialization/serialization.rs
```diff
@@ -0,0 +1,412 @@
...
+
+                h.set_count_at_index(dest_index, count)
+                    .map_err(|_| DeserializeError::EncodedArrayTooLong)?;
+
+                dest_index = dest_index.checked_add(1)
+                    .ok_or(DeserializeError::UsizeTypeTooSmall)?;
+            }
+        }
+
+        // TODO destat is expensive; should accumulate the necessary state while deserializing
```
jonhoo: `restat`?

src/serialization/serialization.rs
```diff
@@ -0,0 +1,412 @@
...
+    let mut bytes_written = 0;
+
+    assert!(index_limit <= h.counts.len());
+
+    while index < index_limit {
+        // index is inside h.counts because of the assert above
+        let count = unsafe { *(h.counts.get_unchecked(index)) };
+        index += 1;
+
+        // Non-negative values are positive counts or 1 zero, negative values are repeated zeros.
```
jonhoo: I'm confused -- how can there be negative values?

src/serialization/serialization.rs
```diff
@@ -0,0 +1,412 @@
...
+/// byte is added to the buf as-is. This limits the input to 8 bytes, but that's all we need.
+/// Returns the number of bytes written (in [1, 9]).
+#[inline]
+fn varint_write(input: u64, buf: &mut [u8]) -> usize {
+    // The loop is unrolled because the special case is awkward to express in a loop, and it
+    // probably makes the branch predictor happier to do it this way.
+    // This way about twice as fast as the other "obvious" approach: a sequence of `if`s to detect
+    // size directly with each branch encoding that number completely and returning.
+
+    if (input >> 7) == 0 {
```
jonhoo: I bit-wise `AND` might actually be faster here.

src/serialization/serialization.rs
```diff
@@ -0,0 +1,412 @@
...
+}
+
+/// Write a number as a LEB128-64b9B little endian base 128 varint to buf. This is not
+/// quite the same as Protobuf's LEB128 as it encodes 64 bit values in a max of 9 bytes, not 10.
+/// The first 8 7-bit chunks are encoded normally (up through the first 7 bytes of input). The last
+/// byte is added to the buf as-is. This limits the input to 8 bytes, but that's all we need.
+/// Returns the number of bytes written (in [1, 9]).
+#[inline]
+fn varint_write(input: u64, buf: &mut [u8]) -> usize {
+    // The loop is unrolled because the special case is awkward to express in a loop, and it
```
jonhoo: Ooof, this is really not pretty. Could we make it more readable with a nested macro instead?

src/serialization/serialization.rs
```diff
@@ -0,0 +1,412 @@
...
+fn varint_read<R: Read>(reader: &mut R) -> io::Result<u64> {
+    let mut b = reader.read_u8()?;
+
+    // take low 7 bits
+    let mut value: u64 = low_7_bits(b);
+
+    if is_high_bit_set(b) {
+        // high bit set, keep reading
+        b = reader.read_u8()?;
+        value |= low_7_bits(b) << 7;
```
jonhoo: Again, macro?

Cargo.toml
marshallpierce: No, I didn't want to conflate the two, but I don't mind merging them

src/serialization/serialization.rs
```diff
@@ -0,0 +1,412 @@
...
+    // This way about twice as fast as the other "obvious" approach: a sequence of `if`s to detect
+    // size directly with each branch encoding that number completely and returning.
+
+    if (input >> 7) == 0 {
+        buf[0] = input as u8;
+        return 1;
+    } else {
+        // set high bit because more bytes are coming, then next 7 bits of value.
+        buf[0] = 0x80 | ((input & 0x7F) as u8);
+        if (input >> 7 * 2) == 0 {
```
jonhoo: Any particular reason this doesn't use `nth_7b_chunk`?

src/lib.rs
marshallpierce: Because that way leads to a whole bunch of `mod.rs`, which I find distasteful. It makes it more difficult or clumsy to glance at a filename and know what it is, or jump to the file you want.

src/serialization/serialization.rs
marshallpierce: Unfortunately that has a performance cost, so until there's a use case I'd prefer to leave it off. I haven't benchmarked it on this particular usecase, but it was nontrivial when I was rewriting the guts of https://github.com/alicemaz/rust-base64.

src/serialization/serialization.rs
marshallpierce: So that the heap-allocated buffers can be re-used across many histograms.

src/lib.rs
marshallpierce: Oops, that's a leftover from an earlier iteration. It isn't needed anymore.

src/serialization/serialization.rs
marshallpierce: We want it to only grow, never shrink. The vec here is a dumb buffer that only exists to be big enough to take an appropriately-sized slice of below. If the buffer is too big here, it might not be the next time...

src/serialization/serialization.rs
marshallpierce: Doh.

src/serialization/serialization.rs
marshallpierce: Maybe, but I'm not at all convinced that this is the final form of the algorithm so I'd prefer to wait until it can't get any faster before we make it pretty.

src/serialization/serialization.rs
marshallpierce: I'll try it.

src/serialization/serialization.rs
```diff
@@ -0,0 +1,412 @@
...
+/// Encode counts array into slice.
+/// The slice must be at least 9 * the number of counts that will be encoded.
+fn encode_counts<T: Counter>(h: &Histogram<T>, buf: &mut [u8]) -> Result<usize, SerializeError> {
+    let index_limit = h.index_for(h.max()) + 1;
+    let mut index = 0;
+    let mut bytes_written = 0;
+
+    assert!(index_limit <= h.counts.len());
+
+    while index < index_limit {
```
jonhoo: I think you could simplify this code somewhat by doing:
```rust
let mut iter = (0..index_limit).into_iter().map(|i|{
    unsafe { *(h.counts.get_unchecked(i)) }
}).peekable();
while let Some(count) = iter.next() {
    // ...
    if count == T::zero() {
        // ...
        while let Some(count) = iter.peek() {
            if count == T::zero() {
                zero_count += 1;
            } else {
                break;
            }
        }
    }
}

File: src/serialization/serialization.rs [hunk 17]
```diff
@@ -0,0 +1,412 @@
...
+
+#[path = "tests.rs"]
+#[cfg(test)]
+mod tests;
+
+#[path = "benchmarks.rs"]
+#[cfg(all(test, feature = "bench_private"))]
+mod benchmarks;
+
+const V2_COOKIE_BASE: u32 = 0x1c849303;
```
jonhoo: Could you separate out `serialize` and `deserialize` into separate files (i.e., `serialize.rs` and `deserialize.rs`), add `mod`s for them here, and then a `pub use serialize::V2Serialize` and `pub use deserialize::Deserializer`.

src/serialization/serialization.rs
```diff
@@ -0,0 +1,412 @@
...
+    }
+}
+
+/// Deserializer for all supported formats.
+/// Deserializers are intended to be re-used for many histograms.
+pub struct Deserializer {
+    payload_buf: Vec<u8>
+}
+
+impl Deserializer {
```
jonhoo: Why is this not `V2Deserializer`?

Cargo.toml
jonhoo: I think merging them is fine.

src/lib.rs
jonhoo: I have not found this to be an issue at all, but fair enough. I don't feel that strongly about it.

src/serialization/serialization.rs
jonhoo: Uhm.. Performance cost in what sense? When constructing the return error?

src/serialization/serialization.rs
marshallpierce: Neat, I'll benchmark it and see.

src/serialization/serialization.rs
marshallpierce: Only the fact that I wrote some of this a year ago as my first Rust project. ;) I have an idea to replace this function with a branch-free version, so hopefully this whole mess can go away momentarily.

src/serialization/serialization.rs
jonhoo: So the use-case is if you want to serialize or deserialize *lots* of histograms right after each other? That doesn't seem all that common? And also comes with an ergonomic penalty.. One alternative would be to instead have `serialize` and `serialize_with`, where the latter takes an `&mut Vec`. This would *allow* users in that particular use-case to have efficient repeated serialization, without incurring the usability penalty for those who just want a simple `to_bytes()` serialization.

src/serialization/serialization.rs
jonhoo: `resize` doesn't deallocate any memory if you set its size to be smaller, it only allocated more if it needs to grow, so you should get the same performance you are after, but with the nice additional benefit that `.len()` is correct (and you can use `[..]` below).

src/serialization/serialization.rs
marshallpierce: I was thinking that the file was getting a little porky, but didn't want to have a disruptive commit while you were reviewing... but now you've given me an excuse. :)

src/serialization/serialization.rs
jonhoo: Having a nested macro would also make experimenting with it easier. You could change the test in just one place, and have it apply to the entire nested expansion.

README.md
```diff
@@ -158,6 +158,45 @@ and this to your crate root:
...
 ```
 
+## Benchmarking
+
+Both Rust's built-in benchmarking and [Criterion](https://github.com/japaric/criterion.rs) are used.
+
+For the benchmarks in `benches`:
+
+```
+rustup run nightly cargo bench
```
jonhoo: I'd be happy to move all the benchmarks to just regular `test` benchmarks. I completely agree with you that the lack of development on Criterion isn't great.

src/serialization/serialization.rs
marshallpierce: I was thinking it would probably be nice to follow in the footsteps of the Java implementation, at least in spirit: there's one deserialize function and it knows how to unpack every flavor. Given the "cookie", it is easy to do the detection for the user.

marshallpierce:
Thanks for your promptness and care doing these reviews! Very pleasant compared to some other projects where work from December is still languishing...

src/lib.rs
```diff
@@ -1394,8 +1403,12 @@ impl<T: Counter> Histogram<T> {
         let mut min_i = None;
         let mut total_count: u64 = 0;
         for i in 0..until {
+            // TODO can panic. May not be a sensible place to use Result, but should audit paths
+            // that can get here.
             let count = self[i];
             if count != T::zero() {
+                // TODO don't unwrap here; weird types may not work.
+                // Fix Counter types to just be u8-64?
```
jonhoo: The only thing I can think of is using `BigInt`, but I suspect that would run into other issues entirely. Limiting to `u8-u64` seems fine, but I'm not sure how we do that nicely (except I guess we could add our own `Num`-like trait and `impl` only for the integer types we want).

src/serialization/serialization.rs
jonhoo: This one seems fine to me. You could alternatively use `Vec::reserve`, but that would still require updating the `len` for every byte you inserted (I think).

src/serialization/serialization.rs
jonhoo: Ohh, I see. Yeah, that seems fine. Maybe add this to the docstring?

src/lib.rs
marshallpierce: Hopefully you'll come to enjoy it as I have (I was introduced to it on other projects I contribute to, and it felt weird to me at first as well). If it really bugs you I won't shed tears if you want it gone later. :)

src/serialization/serialization.rs
jonhoo: Actually, as written it's not quite right. You need to `.peek()`, and then call `.next()` if you actually end up taking the value, otherwise the iterator won't progress.

src/serialization/serialization.rs
marshallpierce: I spent some time digging in to it and between my `perf`-gazing and hassling the LLVM folks, the best I could come up with is that the increased size of the struct affected on-stack layout in those tight loops. It wasn't a huge cost, like 15% or something, but it was easily measurable, regardless of whether or not those errors ever occurred in practice.

src/serialization/serialization.rs
marshallpierce: Actually, repeated use is pretty much the default use-case I have: a bunch of histograms that all need to be periodically serialized and sent elsewhere for an analysis. I maintain https://bitbucket.org/marshallpierce/jersey2-metrics, for instance, which will keep a HdrHistogram for service time of each HTTP endpoint you have. That's the sort of thing I plan to do in Rust, as well. Consider https://github.com/Searchlight/khronus as well: a TSDB built on HdrHistogram. I think we should be viewing more problems through the lens of granular histograms that get aggregated or otherwise combined elsewhere for analysis.

Also, serialization is not a single concept. We already have four distinct version, of which one is implemented and the another will be soon. Would we have bring-your-own-vec versions of both, leading to 4 methods on `Histogram`? That feels sloppy to me. 

From a conceptual standpoint, I think serialization is a fundamentally distinct concern from recording. I think it is a mistake to put `save_to_db()` on your domain model objects/structs, and this feels like an analogous situation. I'd rather have people think about and decide what serialization they want than just see a `serialize()` and hope for the best, and I think an extra line is a reasonable syntactic burden.

src/lib.rs
jonhoo: My editor always shows me the full path to the open file, and I use full-path fuzzy searching to open files, so it doesn't matter whether the distinguishing part of the name is the directory or the filename to me ;)

src/serialization/serialization.rs
jonhoo: Huh, interesting. I'd be great if you could leave a comment about this in the `impl From<std::io::Error>`, and maybe also a link to the discussion?

src/serialization/serialization.rs
jonhoo: Interesting... I was thinking of having a per-version trait like:
```rust
pub trait V2Serialize {
    fn to_bytes(&self) -> Result>Vec<u8>>;
    fn serialize_with(&self, &mut Vec<u8>) -> Result<()>;
}
```
Now, we could implement `V2Serialize`, `V2CompressedSerialize`, etc. for `Histogram`, and users can choose which one they want depending on which trait they import:
```rust
fn main() {
    let h: Histogram<_> = ...;
    {
        use hdrsample::serialize::V2Serialize;
        h.to_bytes(); // will use V2
    }
    {
        use hdrsample::serialize::V2CompressedSerialize;
        h.to_bytes(); // will use V2 compressed
    }
}
```
This gives a consistent API across the different versions, and there's no need for keeping this extra struct around. It is also ergonomic (if you only ever want `V2Serialize`, you just `use` that at the top of your file), and fits well within idiomatic Rust.

I agree with you that people should think about serialization, and I think this API caters to that, but it is *also* easy to use for those who *don't* have a use-case quite as critical as yours.

Cargo.toml
marshallpierce: Argh, it doesn't work, because then it tries to use criterion and can't find the dependency. :(

jonhoo: Maybe we should just get rid of Criterion entirely while we're at it?

jonhoo:
Ouch -- the split made all my comments appear as outdated :(

src/lib.rs
marshallpierce: Yeah, I think the thing that bugs me the most is seeing a list of tabs:

`[mod.rs] [mod.rs] [mod.rs] [mod.rs]`

That and Cargo dependency handling are the places where I feel like the Rust community is backwards. :(

marshallpierce:
Yeah, that's annoying. I'll go through the outdated comments and make sure I've addressed them before I wrap this up.

That, among other reasons, is why I use BitBucket for most of my own work... "Tasks" in reviews are great, and I can't believe GH didn't incorporate them when they did their recent review overhaul. Still not as good as what BitBucket had years ago.

src/lib.rs
marshallpierce: OK, filed https://github.com/jonhoo/hdrsample/issues/29.

README.md
marshallpierce: Filed https://github.com/jonhoo/hdrsample/issues/30.

src/lib.rs
jonhoo: Just don't use tabs ;)

Cargo dependency handling? Really? I've been quite happy with that..

src/serialization/serialization.rs
marshallpierce: Ah, interesting. I'll give it a try.

src/lib.rs
marshallpierce: It's workable, but it repeats the mistakes of older tools when there are better options out there. Not a deal breaker, just a shame...

src/serialization/serialization.rs
marshallpierce: Right but I don't want to have to resize it bigger again when I want it bigger. I want this to be a grow-only buffer, which I take an appropriately-sized slice of as needed.

src/serialization/serialization.rs
marshallpierce: Assuming it stays in nested form, which I hope it won't. :)

src/serialization/serialization.rs
jonhoo: That's what you'll get with `resize` :)
It never shrinks the allocated memory (i.e., the `capacity`), it only grows it.

marshallpierce: (For context on my aversion to repeated `resize` -- see the resize-avoidance in the serializer which netted 40% performance gain!)

andrewjstone:
Wow quick work @marshallpierce and quick review @jonhoo. Thanks! One thing I'd still like to see is support for serialize/deserialize and encodable/decodable. I think that just entails something like this:

```
impl Encodable for Histogram {
    fn encode<S: Encoder>(&self, s: &mut S) -> Result<(), S::Error> {
       let mut s = V2Serializer::new();
       let mut vec = Vec::new();
       let _ = s.serialize(&h, &mut vec).unwrap();
        s.emit_struct("Histogram", 1, |s| {
            s.emit_struct_field("data", 0, |s| {
                s.emit_from_vec(&vec, |s,e| {
                    s.emit_u8(*e)
                })
            })
        })
    }
}
```

We'd then do the converse for decoding. Make sense? I can take a stab if you want as it doesn't necessarily have to be in this PR. Or you can feel free to do it :)

marshallpierce:
I don't think that sort of "default encoding" is a good concept to give to users. That sort of thing feels attractive (see: built in serialization in tons of languages), but it leads to tremendous long term pain with versioning, or when there are multiple viable options (see: how horrible built in serialization has turned out to be in basically every case). Why would we choose, for instance, to use V2 rather than V2 + `DEFLATE` compression? The Java impl has both, and both are in use in the wild. There is not just one flavor of compression, and users have to choose what makes sense for them and their environment. Maybe they have an old histogram processing pipeline that only supports V1 or the ancient V0 encoding (in which case, file an issue!). 

What about when V3 comes out? We can't change what `encode` means because that would be sneaky and users would be justifiably upset, but we also wouldn't want to leave it using the old version... To create any sort of implicit or explicit default like this is laying a trap for ourselves when inevitable change occurs.

Unfortunately, it's not (yet?) really feasible for users to make their own Encodable implementations either because of coherence rules, but I strongly feel it would be a mistake to bless any particular serialization scheme as the one that `encode` would perform.

What is the broader use case you're trying to achieve with Encodable?

jonhoo:
@marshallpierce I think it would be really good to at least have `impl`s for `serde::Serialize` and `serde::Deserialize`, as many libraries require those types to do persistence and rpc of custom types. I am sympathetic to your argument that people should think carefully about serialization, but I do think that there are a non-trivial number of users who will just want to use `Histogram` for recording a bunch of stats and then send them over, say, RPC, with no further fuzz. Since the `HdrHistogram` serializer is (at least in theory) common across all version formats, I think we actually could upgrade the default serialization under the hood.

An alternative design which you might like better is to provide wrapper structs for serialization. Something like `wrappers::v2::Serialize : From<Histogram>` which has `impl Serialize` and `impl Deserialize`. It's less ergonomic, but would give the control you're after..

andrewjstone:
My use case is a location transparent actor system, [rabble](https://github.com/andrewjstone/rabble). The inter-node communication is performed using [msgpack](https://github.com/3Hren/msgpack-rust) which requires the user [type being serialized to be Encodable](https://github.com/andrewjstone/rabble/blob/master/src/msg.rs). It's very useful for users of the library to simply `#[derive(RustcEncodable)]` for an arbitrary type. They don't care how the types are encoded, they just want to ship them between nodes. This is very similar to how things work in Erlang, except Erlang automatically serializes all types for internode communication. Users can still version those types in the types themselves in case the structure of the type changes in a long lived system. They care about the versions of the objects, not the versions of the serialization. 

Implementing Encodable/Decodable allows the library itself to swap msgpack for protobuf or some other serialization format in the future without having to change the user code. Versioning of the protocol and inter-node negotiation would be needed by the library, but the user won't have to worry about that. They just upgrade each node in a rolling fashion. 

Now, this is different from your concern, where you are worried about the type not changing, but want to be able to serialize the same type using new (more efficient) serialization that may exist in the future. This is a legitimate concern that I didn't think about. I think it may be possible to include all types of serialization in a manner similar to what I laid out in the other post. We could do something like the following where we version the serialization struct, and check a config variable (via feature flag) to see which format to serialize to. For deserialization, we check the version in the serialized version and deserialize using that version of the serialization code. The code always defaults to using the latest version.

```rust
impl Encodable for Histogram {
    fn encode<S: Encoder>(&self, s: &mut S) -> Result<(), S::Error> {
        let (version, serializer, vec) = if cfg!(v2_serialize) {
            let mut serializer = V2Serializer::new();
            let mut vec = Vec::new();
            (2, serializer, vec)
        } else {
          ...
        };
        let _ = serializer.serialize(&self, &mut vec).unwrap();
        s.emit_struct("Histogram", 2, |s| {
            try!(s.emit_struct_field("version", 1, |s| {
                s.emit_u8(version)
            }));
            s.emit_struct_field("data", 0, |s| {
                s.emit_from_vec(&vec, |s,e| {
                    s.emit_u8(*e)
                })
            })
        })
    }
}
```

There would need to be documentation explaining the versioning, but other than that I think it solves the problem with minimal work for the user. Of course, old versions of the code won't be able to deserialize newer versions, but that's true of everything. Backwards compatibility will be maintained.

With all that being said, I don't believe coherence rules are an issue necessarily, since a wrapper type could be created that does the same thing, but just picks a serialization version. The user can then upgrade or maintain backwards compatibility at their own whim. However, this requires more work for the average user of the library that just wants to serialize some histograms in msgpack, etc... and doesn't care what version of serialization is used.

andrewjstone:
I just want to add, that for my use case I could totally implement the wrapper type and users wouldn't have to worry about it, since I intend histograms to be built in as metrics. However, I still think it would be easier for most users of hdrsample to have a ready made implementation for them.

marshallpierce:
I agree that there is a persistent user desire to have serialization "just work". Unfortunately, there is a similarly long history of such efforts working out poorly in practice. 

Let's work through a couple of possibilities with a hypothetical project Foo that uses this library.

If we implement `Serialize` or `Encodable` or any other effort to achieve "standardized" serialization, and we change from V2 to V3, or V2 + DEFLATE in some future update, we break all consumers of those serialized histograms. It is not a real-world realistic goal to think that all consumers of Foo's output are going to be upgraded with new versions that can deserialize V3 histograms before Foo updates its dependency version. There maybe certain situations where small systems are deployed in lock step, or something like that, but there are decades of history showing that flag days do not work in practice, so we should not trap our users into requiring one.

If we *don'*t update in future versions, then we trap all users on the current format forever, which is also clearly wrong.

The problem here is the very concept of "standardized serialization". Attempting to have each type know how to serialize itself is possible, but not in any broadly useful way. The software community has already tried this. CORBA. Java RMI. Python pickling. Ruby marshaling. It's incredibly alluring but it _does not work in practice_. Serialization is an inherently contextual thing, and forcing library authors to try to solve the problem once and for all is fundamentally impossible to do well. If we only ever get one way to expose a serialization API, then we effectively can't change it. But, it's also dreadful to not change it. There's no right answer, because concerns about whether or not a given process needs to be able to serialize or deserialize a given version or format are fundamentally application specific.

This is a perfect example of the problem. @andrewjstone is using a framework that has fallen for an alluringly "simple" solution to the serialization problem: just have every type know how to serialize itself, in the "one best way", whatever that is! It will make it so easy to make communication over the network transparent. Oh wait, serialization is more complex than that. Versioning. Compression. This has all been done before: Java RMI _is exactly this concept_ and _it was awful_. It was extremely popular for a short while a couple decadess ago because it was so fast and easy to get started with, and then got extremely unpopular forever after because it turned out to be untenable in practice to have core parts of your system rely on fairly opaque and brittle magic serialization. (It, and all similar "magical" serialization tools, have turned out to be a rich source of security vulnerabilities over the years as well.) 

The reason why these messy wrapper structs are what we're left with is because serialization-via-standardized-trait is the wrong approach. You can only implement a trait once for a struct, and I hope I've at least given a glimpse of why that's a bad idea for serialization. Thus, we have wrapper structs, which feels bad because it is bad. This is better solved with a visitor pattern approach. Different serialization? Different visitor. Upgrading serialization version? Easy, and in the application's control. Perhaps more importantly, *not* upgrading the serialization version accidentally? Also easy. In pattern-speak, the standardized trait approach is a pretty nasty violation of [SOLID](https://yow.eventer.com/yow-2013-1080/the-solid-design-principles-deconstructed-by-kevlin-henney-1386). In general, coupling persistence to domain models has been tried again and again over the years; every platform seems to reinvent this particular problem, and I really wish Rust wasn't repeating that mistake. It just doesn't work well.

If you want to use a trait in your app to represent serialization, go right ahead (though I think you'll regret it in the end), but at least make it your own trait, not a trait that hands responsibility for serialization and all its complexity off to someone else and doesn't give you design flexibility. We could provide wrapper structs for Encodable, and Serialize, and anything else, but I think we should not be making it easier for people to participate in an ecosystem antipattern. If they want them, they are trivial to write around the existing serializer/deserializer structs.

Whew! End rant. :) I really don't enjoy raining on your respective parades, but this is a _bad idea_ and I will do my best to steer Rust away from repeating 20 year old mistakes, even if I only convince you two...

jonhoo:
> Whew! End rant. :) I really don't enjoy raining on your respective parades, but this is a *bad idea* and I will do my best to steer Rust away from repeating 20 year old mistakes, even if I only convince you two...

I have no problem with rants, especially when they are productive and well-informed :)

> If we implement `Serialize` or `Encodable` or any other effort to achieve "standardized" serialization, and we change from V2 to V3, or V2 + DEFLATE in some future update, we break all consumers of those serialized histograms. 

Ah, so you're worried about *forward* compatibility specifically? About senders upgrading while the receiver does not. The way I see it, switching major versions when the default serialization format changes, and being clear about forwards-compatibility errors when decoding is a perfectly reasonably solution to this *for the common user*. Telling them that they cannot use `Histogram` with *any* of the existing Rust frameworks for persistence, RPC, and automatic distribution without writing custom (de)serialization code themselves that wrap a particular version of the serialization format seems like a major disadvantage that many would find hard to swallow.

It is not as though we are *precluding* users from using a particular (explicit) serialization format. My proposal is certainly not to remove the versioned serialization methods, or anything along those lines. My proposal is to pick the most common one, and provide that as the default `Serialize` implementation (`Deserialize` has no such "picking" since we can determine based on the version cookie). This can be overridden by the user if they need to talk to legacy systems, or if they want to upgrade before the default is changed.

> The problem here is the very concept of "standardized serialization". Attempting to have each type know how to serialize itself is possible, but not in any broadly useful way. 

I don't personally agree on this point. I would argue that both XML, and lately protobufs, have been *extremely* successful in doing this. Now, that's not to say that it makes sense in all cases, and I completely agree with you that users should have the *option* to explicitly choose their serialization. But I still think we should make the common case easy.

> Oh wait, serialization is more complex than that. Versioning. Compression.

I disagree with this point. Here, multiple layers of the transport stack have proven a robust and efficient solution. Look at the web today — compression and encryption are handled entirely by the transport layers, and website authors need to concern themselves with it at all. Similarly, browsers treat those simply as layers to be peeled off before handing the raw HTML to the parser and renderer.

> You can only implement a trait once for a struct, and I hope I've at least given a glimpse of why that's a bad idea for serialization.

I think this is the crux of our disagreement, and I don't think we even disagree! There is *no* implementation of `Serialize` that works for everyone. And we shouldn't *force* users to stick to one and only one serialization format. I would never argue for that. What I *do* argue is that there *should* be an implementation of the common serialization pattern that people can use *if it doesn't matter to them*. A default for the 80% use-case.

> This is better solved with a visitor pattern approach. Different serialization? Different visitor.

This is exactly what serde uses, and how you implement `Serialize` and `Deserialize` for custom types. Using that approach alone does not solve the issue of letting users *choose* their serialization though, except by saying "you need to implement it from scratch yourself". But that's sort of beside the point.

> In general, coupling persistence to domain models has been tried again and again over the years

Again, this is not what I am advocating for. I want us to provide multiple serializers, and let the user choose among them. **But**, I want us to have an implementation of the "standard" serialization trait (as bad of a pattern as it may be) where *we* pick one of our serialization implementations that we think caters to the most common use-case. If someone has a different use-case, they can with relatively little effort choose to use a different serializer that we provide, but if their use-case matches the one we cater for, we have reduced the cognitive overhead required for them to use `HdrHistogram`.

-----------------------------------------------------------------

All that said, here's what I think we should do:

 - Have multiple different traits, `serialization::v2::Serialize`, `serialization::v2_compressed::Serialize`, etc. (and symmetrically for `Deserialize`), each of which define exactly two methods: `to_bytes(&self) -> Vec<u8>` and `serialize_with(&self, &mut Vec<u8>)`. By keeping the same API between the different versions, users can choose (and *swap*) between formats easily (just by choosing which trait to `use`), and the serialization story gets easier to explain.
 - Implement said traits for `Histogram`
 - Implement `serde::Serialize` for `Histogram`, with an implementation that simply calls `serialization::v2::Serialize::to_bytes` (and similarly for `Deserialize`, though its call is version-independent)
 - In the docs have a section on serialization that says: "`Histogram` is a complex datastructure whose serialization is not straightforward. While we provide an implementation of standard serialization traits for convenience, more heavy-weight users should explicitly choose to use one of our serialization strategies from `serialization::*` in their encompassing `Deserialize` implementation".
 - The release notes for any major version bump should clearly state if the `Serialize` implementation has been upgraded (say V2 -> V3), give code for how to continue using the old format (which shouldn't be too hard), and warn that continuing to use `Serialize` directly will require upgrading the receiver as well. I am perfectly okay with this, as a major version bump signifies a *backwards incompatible change*.

marshallpierce:
> Telling them that they cannot use Histogram with any of the existing Rust frameworks for persistence, RPC, and automatic distribution without writing custom (de)serialization code themselves that wrap a particular version of the serialization format seems like a major disadvantage that many would find hard to swallow.

Really? I guess I just don't find that plausible. `V2Serializer::new().serialize(h, &mut writer)` doesn't seem like much to ask them to write. We give them a menu of choices that trivially provide the ability to write to a `Write`, and they can't figure out how to glue that together in a way that makes sense in their environment? Anyway, I don't want to lose focus by debating this particular point because whether we provide trivial impls for each version or they write them is not the core decision.

> It is not as though we are precluding users from using a particular (explicit) serialization format. 

I don't understand how we are _not_ doing that if we do things like implement `Encodable` or `Serialize`.

> I don't personally agree on this point. I would argue that both XML, and lately protobufs, have been extremely successful in doing this. Now, that's not to say that it makes sense in all cases, and I completely agree with you that users should have the option to explicitly choose their serialization. But I still think we should make the common case easy.

... [snip]  ...

> Look at the web today — compression and encryption are handled entirely by the transport layers, and website authors need to concern themselves with it at all. Similarly, browsers treat those simply as layers to be peeled off before handing the raw HTML to the parser and renderer.

I don't think the comparison applies here because this isn't about layering. XML has done nothing to enable one type of serialization. It is a way to prevent people from needing to learn `yacc` to parse complex text, nothing more. This isn't about XML; the analog to XML in this scenario is bytes: a low level primitive that can be used to build a format. This _is_ about magically serializing in-memory structures to XML, or JSON -- XML or JSON or protobuf or Cap'n Proto or are merely means to an end, and *you have to decide* how to use them. There have been tools to turn object graphs into XML and vice versa, and they are good for tech demos and quick hacks and little else because they are intolerably fragile. I would not wish maintaining a service that relies heavily on JAXB reflecting into domain objects on the meanest soul. Just saying you use XML gets you very little in terms of interop, only the ability to have a tool do some lexing for you. If that worked, we'd all have stopped with UDDI and serialization (and much of business integration, if you believe the UDDI hype) would be a solved problem... 

> What I do argue is that there should be an implementation of the common serialization pattern that people can use if it doesn't matter to them. A default for the 80% use-case.

Even though in the long term it will bring pain to whatever system uses the default, when the default can and will change? By coupling things like default serialization changes to version upgrades, we increase the risk of and aversion for dependency updates in our consumers. This is a problem that doesn't need to exist. It's one thing to provide something like `new_with_max` that sets the lower bound to 1. That's a pretty defensible thing to do, and there isn't an implicit concept of versioning of 1s in the world. Without a reworking of the histogram format, 1 is it. Serialization is much more fluid than the fact that 1 is the lowest value positive value.

> This is exactly what serde uses, and how you implement Serialize and Deserialize for custom types. Using that approach alone does not solve the issue of letting users choose their serialization though, except by saying "you need to implement it from scratch yourself".

I guess I'm misunderstanding serde, then. Based on https://serde.rs/impl-serialize.html, it looks like I get one chance to implement it for a given type. 

```
impl Serialize for i32 {
...
}

```

That doesn't look very visitor-y, choose-your-flavor to me. It looks like I get 1 implementation of `Serialize` ever for a given type, not a choice among external "serializers" that navigate along some graph and serialize as they go. What am I missing?

> I want us to provide multiple serializers, and let the user choose among them. But, I want us to have an implementation of the "standard" serialization trait (as bad of a pattern as it may be) where we pick one of our serialization implementations that we think caters to the most common use-case. If someone has a different use-case, they can with relatively little effort choose to use a different serializer that we provide, but if their use-case matches the one we cater for, we have reduced the cognitive overhead required for them to use HdrHistogram.

I guess I just don't agree with that philosophy. I don't want to reduce cognitive overhead by papering over a problem that will bite them later. I'd rather say "Here are your options. You should probably pick X unless you have other requirements, but they all will output to a `Write` so you can plug them in to basically anything." I hate to sound like a grumpy old man, but if that is too much of a burden, then I'm not sure how to help in that scenario, but I'm pretty confident it's not just ignoring the issue.

> Have multiple different traits, serialization::v2::Serialize, serialization::v2_compressed::Serialize, etc. (and symmetrically for Deserialize), each of which define exactly two methods: to_bytes(&self) -> Vec<u8> and serialize_with(&self, &mut Vec<u8>). By keeping the same API between the different versions, users can choose (and swap) between formats easily (just by choosing which trait to use), and the serialization story gets easier to explain.

If I'm understanding you correctly, this is an odd beast type-wise: traits that all have methods that are spelled the same way, but are not actually type-system interchangeable ala polymorphism? This is crying out for polymorphism, so let's just do that instead: define a `HistogramSerializer` and let you select which struct (V2, V2 + DEFLATE, etc) to use for it. Pluggability of serialization is orthogonal to the `Histogram` itself; _it should be a separate concept_. The fact that trying to bolt this into the `Histogram` itself leads towards this awkward "polymorphism via spelling" construct could be a signal that there is some sort of square peg/round hole situation...

> The release notes for any major version bump should clearly state if the Serialize implementation has been upgraded (say V2 -> V3), give code for how to continue using the old format (which shouldn't be too hard), and warn that continuing to use Serialize directly will require upgrading the receiver as well. I am perfectly okay with this, as a major version bump signifies a backwards incompatible change.

Again, this is completely needless complexity to inflict on the user. They have a hard enough time dealing with version upgrades as it is. A version upgrade where the same API produces materially different results when that same difference was both predictable and easily avoidable? What would you think about some other library that did that?

My alternate proposal:

- Have structs for each of the serializer formats. V2, V2+DEFLATE, etc. that are as low-level and widely applicable as we can make them. (I'm not sure we can do any better than writing to `Write`? That can be a `Vec` or slice or an i/o stream, and I think that pretty much covers the bases.)
- Perhaps have them all implement a common `HistogramSerializer`, because that has negligible architectural cost, and maybe someone would want to polymorphically choose between them.
- If users want to have a `Histogram` be `Encodable`, they should wrap it in 1-element struct so that they have an opportunity to implement `Encodable` the way they see fit: V2, JSON, whatever. Such an implementation is 1 line if they want to use our serializers: `SerializerOfChoice::new().serialize(histogram, writer)`. 
- If I'm misunderstanding serde and there is pluggability that I'm missing, I think it's fine to provide implementations of that for each of the serializer options and let the users choose, with documentation guiding them to a sane default.
- When we ship new versions, all user code keeps working until they explicitly change serialization versions, forever.

andrewjstone:
>  It is not a real-world realistic goal to think that all consumers of Foo's output are going to be upgraded with new versions that can deserialize V3 histograms before Foo updates its dependency version. There maybe certain situations where small systems are deployed in lock step, or something like that, but there are decades of history showing that flag days do not work in practice, so we should not trap our users into requiring one.

I agree it's unrealistic to expect all nodes in a system to update at once. I explicitly stated upgrades should be in a rolling fashion. Regardless of implementation, both backward and forward compatibility must be taken into account.

> This is a perfect example of the problem. @andrewjstone is using a framework that has fallen for an alluringly "simple" solution to the serialization problem: just have every type know how to serialize itself, in the "one best way", whatever that is! It will make it so easy to make communication over the network transparent. Oh wait, serialization is more complex than that. Versioning. Compression. This has all been done before: Java RMI is exactly this concept and it was awful. 

What I'm actually trying to accomplish is the opposite. I want the serialization to be selectable by the library (rabble). The library is effectively a message passing runtime. All interactions in the systems are performed via message passing. Messages get passed to processes with a named pids. Communication between pids on the same node don't need serialization. When sending a message from node to node, the types need serialization. The specific format of serialization is up to rabble. The types only implement the visitor pattern using the Serialize/Deserialize traits, so that the library can serialize them and pass them between nodes. This is a layered system, where internode communication and versioning for forward and backward compatibility between nodes is done by the library. When a node connects to another node there is a handshake to determine protocol versions etc... Users should never even have to know how their types are serialized. The only reason this has become a problem and I believe has raised your concerns is because hdrhistogram is explicitly picking a serialization format and saying "hey, you have to use our serialization format to get things done effficiently". And unfotunately hdrhistogram doesn't have just one, so building a system requires users to now explicitly choose their serialization formats for a single type. This is what you have called "having a type know how to serialize itself". It makes a lot of sense for hdrhistogram, which is why it's done this way. Unfortunately, only histograms can be serialized in this manner. So for a message passing system, there needs to be a wrapper type that can use the best serialization at the time for the specific type, but still allow the library to wrap those serialized messages, so that it can send them across nodes via the internode protocol. You don't want each type having it's own serialization and then having to decipher each individual type in a type specific manner at each node endpoint. You want uniformity, which is what the Serialize/Deserialize traits combined with a versioned protocol and serialization format give you.

 My proposal was simply to take the best of both worlds. Allow type specific specialization using hdrhistogram, then take the already serialized type and embed it in the visitor pattern as a Vec<u8> so it can be used with other libraries like rabble. The core issue here is now users can't just ignore the inter-node protocol in rabble changing, since they may want to upgrade hdrhistogram and use a newer serialization format for better performance (for histograms only). At this point, version negotiation can no longer be done by rabble nodes and you end up specifically with the problem you describe of being unable to perform a rolling upgrade to change histogram serialization versions. Once again, this is a very great observation, @marshallpierce and I thank you for brining it to my attention. I think you'll be surprised at my conclusion, which is that now I think it is indeed *incorrect* to provide a general Serializable/Deserializable trait precisely because hdrhistogram has a very specific serialization format and users need to decide which version they use. Rabble should implement serializable in a versioned wrapper type specific to the serialization, so that the clustering code can deal with interchange between nodes that haven't upgraded to the new version yet. That being said, I completely disagree with Serializable/Deserializable being a bad pattern in general and have found it quite useful, especially for user defined arbitrary types. 

Now, regarding RMI, I agree that it was terrible, along with it's predecessor, RPC. There are many problems there in addition to versioning and serialization. I'll go so far as to say that the reason people hate RPC has absolutely nothing to do with versioning or serialization. The RPC runtime can easily manage version negotiation. The core problem is that RPC/RMI treats remote requests like local ones. This ignores the asynchrony and partial failure inherent in distributed systems. Actor systems, like erlang and Akka do not have this problem. They take the opposite approach and treat all operations as remote via message passing. All operations can fail, messages can be dropped etc.. Erlang is well known for building the most reliable systems ever, built with non-stop operation. Now, there is more to this than just  message passing, including dynamic code upgrade of running systems, but that is neither here nor there. The fact of the matter is that erlang has a standardized inter-node serialization format that users of the language don't have to know anything about. This format has been upgraded and added to over the years, and version negotiation between nodes at runtime allows for both forward and backward compatibility via rolling upgrade of one node at a time. 

> You can only implement a trait once for a struct, and I hope I've at least given a glimpse of why that's a bad idea for serialization. Thus, we have wrapper structs, which feels bad because it is bad. This is better solved with a visitor pattern approach. Different serialization? Different visitor. Upgrading serialization version? Easy, and in the application's control. 

As @jonhoo has pointed out the Serialize/Deserialize traits do implement the visitor pattern. They don't perform any actual serialization. That is up to the serializing crates like messagepack, protobuf, capn-proto etc... 

> If you want to use a trait in your app to represent serialization, go right ahead (though I think you'll regret it in the end), but at least make it your own trait, not a trait that hands responsibility for serialization and all its complexity off to someone else and doesn't give you design flexibility.

I'd much prefer to stick with standardized types that implement the visitor pattern so that I don't have to re-implement msgpack or protobof again for no good reason.

> Whew! End rant. :) I really don't enjoy raining on your respective parades, but this is a bad idea and I will do my best to steer Rust away from repeating 20 year old mistakes, even if I only convince you two...

I quite enjoyed the rant, and it made me rethink a few things: particularly, that my solution above with versioning embedded in the Serialize/Deserialize Impls is wrong. I hope my counterpoints, and conclusion can show you that I understand the fallacies of distributed systems and that I'm trying to build an actor system in a principled manner that is also ergonomic.

Now onto @jonhoo's proposal. I think I may have been convinced that it's incorrect in this case, as with my proposal. Outside of actor systems, serialization is mainly used for file storage/persistence and client/server interfaces. Persistence is a single node issue, and versioning is trivial. When the code on that specific computer is upgraded it should either read old data with the old code, and read and write new data with the new code or have an ETL step that reformats the old data to the new serialized format. Either way, the code needs to support multiple versions of serialization at a time, and a single Serializable trait makes that more difficult and tricks the user into thinking it's a solved problem. The only way the Serialize impl makes sense is if the ETL step is used, as you can't use multiple versions of the serialized code otherwise.

For client/server systems the issue is a tad more complicated. An upgraded server will have to handle all versions of serialization and clients can upgrade over time subject to deprecation of older versions. After some point the clients will receive versioning errors. Clients can also utilize old and new versions for servers that have not yet upgraded and follow a similar deprecation path. In this case I think the Serialize impl does a disservice to application devs as @marshallpierce pointed out. 

I think the opitimal solution is to have a different histogram type for each serialization version or just leave things as is and let the user implement the wrapper and worry about versioning. It's not much work at all, and prevents making hard things appear easy.

andrewjstone:
At this point I think @marshallpierce and I agree. He has convinced me that Serialize/Deserialize being implemented for Histogram is a bad abstraction. I think where we disagree is in implementing those traits for general user defined types. Most user defined types are simply structs that don't have the low level complexity/perf characteristics of a Histogram. They can be serialized in a general manner and contain a version field for application level negotiation. Things like Histograms, Video files, etc.. are much more complicated, and generalized serialization for formats such as those is likely going down the wrong path. Serialize/Deserialize must be implemented only on versioned wrapper types in that case.

marshallpierce:
@andrewjstone My apologies, I wasn't intending to malign you or your system; I was referring to the requirement of msgpack that things implement Encodable. I should have been more clear.

A couple more points:

As far as I can tell, the form of "visitor" embodied by `Serialize` is pretty weak. There can be different `Serializer` implementations passed to `serialize`, but since you can only `serizlize` once, it looks like the implmentor has to pick that their struct will be, say, two u32s and a bool. That could end up in JSON or Protobuf but it will always be two u32s and a bool. This is a much less generic solution than what I think the visitor pattern is generally used for: it basically allows the user to pick CSV or TSV for a given set of quanta of data rather than performing arbitrarily complex serialization decisions.

A visitor pattern is often used to implement things like crawling a compiler's IR. You might have a pretty-printing visitor for human consumption, and a peephole optimization visitor to turn `x / 2` into `x >> 1`, and an x86 assembly visitor. It's up to the visitor what it's going to do at each node. The visitor isn't told "two u32s and a bool" and expected to figure it out from there, it introspects at each point and does implementation-specific logic (which might not be serialiation at all). 

Now imagine a world where serde was there to help you write this latter form of visitor. Your serialization logic is now neatly partitioned, and easily pluggable. Want a different serializer for `Foo`? Plug a different implementation in when you apply your visitors. It would be a much better sign if there was a trait like `ThingSerializer<T>` that you would make `impl<Foo>`s of for `FooSerializerV1`, `FooSerializerV2`, etc. Then you wouldn't need to mangle your object graph to insert those concepts. Instead, you could apply versioning as a separate concern: choose your FooSerializerV2 today and V3 tomorrow without having to insert a bunch of wrapper structs. Basically, the "serialize yourself" approach forces you to conflate object graph relationships with versioning because it gives you no other choice: the only tool you have is wrapping. It sounds like this is (hypothetical) approach is a better match to your desired usecase. 

As an example, consider a popular tool for JSON serialization in the JVM world, Jackson. Rather than implementing an interface (trait for us here) like `Jsonable` or whatever, Jackson crawls around your object graph and writes data based on just default rules (for simple cases), annotations (for more customization), or delegation to a custom serializer for a given type (for when you need to get really wild). I'm not saying that Jackson is perfect, but I _am_ saying that this approach has proven to be flexible and unobtrusive precisely because of the attributes we've already belabored, and I wish that serde was going in that direction instead of the one they are following. This way, the simple cases are still easy, but you have complete control over changing serialization behavior in a way that's orthogonal to things like upgrading library dependencies because that concern has been split apart.

Even with tools like Jackson, or more primitive automagical serializers like the aforementioned JAXB, it's best practice to create simple dummy classes (in JVM parlance) that you can use to express versions. If you were writing a web service that hosted pictures of kittens, you might have `V1KittenMetadata` and give it the right fields (perhaps even relying on Jackson's default field name to JSON conversion, and just never changing the field names) and replacing it with `V2KittenMetadata` when the time came, providing a custom deserializer if you needed to in order to deserialize `V1`'s JSON. These classes do not have business logic in them; they are just a bag of fields and that's it. So, there can still be a role for what we would need to do with wrapper structs here; they just don't *have* to be wrappers. They can be more like customized views into certain underlying data without needing to interpose in the object graph.

src/serialization/serialization.rs
```diff
@@ -0,0 +1,24 @@
...
+#[cfg(test)]
+mod tests;
+
+#[path = "benchmarks.rs"]
+#[cfg(all(test, feature = "bench_private"))]
+mod benchmarks;
+
+#[path = "v2_serializer.rs"]
+mod v2_serializer;
+pub use self::v2_serializer::{V2Serializer, SerializeError};
```
jonhoo: Either `SerializeError` should be placed directly in this file (if it is common to all serializer versions), or it should be renamed to `V2SerializeError`

jonhoo:
While I'm still not convinced that `Serialize`/`Deserialize` is in and of itself an anti-pattern, I think you've convinced me that the overhead for a user to implement those traits themselves using a wrapper is fairly small, and therefore that's the approach we should take.

In terms of what remains for this PR, I think the biggest thing I would like to see if we're not implementing `Serialize` and `Deserialize`, is improved documentation for the `serialization` module. In particular, I'd like to see an "examples" section which shows both how to serialize + deserialize a histogram, how to deserialize and merge, and how to implement `Serialize`/`Deserialize` with a wrapper struct. I think that would be extremely helpful to someone who wants to start using this crate, and also needs serialization.

Separately, I'd also like the documentation for `serialize` and `deserialize` to point out that `Vec<u8>` implements `Write`, and that `&[u8]` implements `Read`. I imagine a non-zero number of users will just want to get the bytes (or deserialize a bunch of bytes) that they have already read or will write later, and will be confused about how to make a `Write` or `Read` as the API requires.

marshallpierce:
Good point; things like that that which might seem clear to me (because I just wrote tests that treat a `Vec` as a `Write` so it's fresh in my mind) would be useful to explain further.

src/serialization/serialization.rs
marshallpierce: Yeah, I think it will need to be `V2SerializerError` unfortunately because it has error cases that, for instance, V0 does not because that didn't have variable-length encoded ints. That may well be the undoing of being able to genericize over serializers; we'll see.

src/serialization/v2_serializer.rs
```diff
@@ -0,0 +1,226 @@
...
+/// byte is added to the buf as-is. This limits the input to 8 bytes, but that's all we need.
+/// Returns the number of bytes written (in [1, 9]).
+#[inline]
+pub fn varint_write(input: u64, buf: &mut [u8]) -> usize {
+    // The loop is unrolled because the special case is awkward to express in a loop, and it
+    // probably makes the branch predictor happier to do it this way.
+    // This way about twice as fast as the other "obvious" approach: a sequence of `if`s to detect
+    // size directly with each branch encoding that number completely and returning.
+
+    // TODO try bitwise and instead of shift
```
marshallpierce: `AND` is slower. Experimenting with hand-inlining all the helper method calls shows that the compiler already does a great job of hoisting out the common sub-expressions, and just the fact that `AND` is a different calculation than the shift (which eventually gets re-used in later branches) means that it ends up worse. I looked up those instructions in the latency tables and for a register and immediate, both `and` and `shr` have latency 1, so this result isn't too surprising. If the shifts weren't already being used, it should have been equivalent.

src/serialization/serialization.rs
marshallpierce: Good news; the cost of including ErrorKind was unmeasurably small, probably because this format is not as optimizable as base64 so it just gets lost in the noise.

src/serialization/serialization.rs
marshallpierce: I couldn't think of something that would be easier to understand than what it is. You're welcome to give it a try, but 3 of the 9 nesting levels are special cases, and this code is basically never going to change again, so I don't think there's much to be gained. It's long, but it is at least eminently comprehensible, and I worry that macros fancy enough to specify the special cases would not be.

andrewjstone:
@marshallpierce 
> @andrewjstone My apologies, I wasn't intending to malign you or your system; I was referring to the requirement of msgpack that things implement Encodable. I should have been more clear.

Ahh, thank you. No offense taken. I was actually explaining to a friend on IRC late last night how much I was enjoying the conversation and thoughtfulness of discussion. I wasn't even necessarily sure that you were talking about rabble, but figured I'd give you an overview anyway.

I also now understand your point about the visitor pattern and what you were trying to get across. This discussion really was enlightening and got me to think about rolling upgrade in rabble and how user message types should be structured to facilitate that. I just had a long chat with a former coworker who started the project that rabble is meant to be used in to discuss rolling upgrade because of this whole serialization debate. We worked out a nice way to do things that is closer to your ideal than what currently exists. Oddly enough, we will not be using mspack or serializable/encodable anymore 😸 

It's going to be quite a bit of churn, but that means a fun weekend. Cheers!

marshallpierce:
Cool, I'm glad I was able to stimulate a good discussion. :) It would be great to be able to whiteboard this sort of thing, but sometimes async communication is the best we can do. Quite a PR this has turned out to be!

src/serialization/serialization.rs
marshallpierce: Alas, it costs about 20%.

Plain indexing:

```
test serialize_large_dense     ... bench:   6,692,299 ns/iter (+/- 285,733)
test serialize_large_sparse    ... bench:   6,187,564 ns/iter (+/- 169,734)
test serialize_medium_dense    ... bench:      38,908 ns/iter (+/- 963)
test serialize_medium_sparse   ... bench:      34,861 ns/iter (+/- 686)
test serialize_small_dense     ... bench:       3,862 ns/iter (+/- 72)
test serialize_small_sparse    ... bench:       1,727 ns/iter (+/- 77)
test serialize_tiny_dense      ... bench:       1,049 ns/iter (+/- 44)
test serialize_tiny_sparse     ... bench:         491 ns/iter (+/- 13)
```

Peek:

```
test serialize_large_dense     ... bench:   8,016,655 ns/iter (+/- 350,130)
test serialize_large_sparse    ... bench:   7,649,271 ns/iter (+/- 248,415)
test serialize_medium_dense    ... bench:      58,660 ns/iter (+/- 1,349)
test serialize_medium_sparse   ... bench:      56,437 ns/iter (+/- 1,489)
test serialize_small_dense     ... bench:       5,655 ns/iter (+/- 158)
test serialize_small_sparse    ... bench:       2,270 ns/iter (+/- 53)
test serialize_tiny_dense      ... bench:       1,108 ns/iter (+/- 33)
test serialize_tiny_sparse     ... bench:         702 ns/iter (+/- 22)
```

marshallpierce:
OK, take a look at that documentation and see what you think. If it looks good, I think this is ready to go. There's still more to be done (deserialize optimization, deflate support, etc) but I think it''s good enough to get into people's hands.

src/serialization/serialization.rs
```diff
@@ -1,3 +1,167 @@
...
+//!
+//! For the V2 format, the space used for a histogram will depend mainly on precision since higher
+//! precision will reduce the extent to which different values are grouped into the same bucket.
+//! Having a large value range (e.g. 1 to `u64::max_value()`) will not directly impact the size if
+//! there are many zero counts as zeros are compressed away.
+//!
+//! # API
+//!
+//! Each serialization format has its own serializer struct, but since each format is reliably
+//! distinguishable from each othere, there is only one `Deserializer` struct that will work for
```
jonhoo: other

jonhoo:
LGTM once that typo is fixed!

src/serialization/serialization.rs
marshallpierce: Doh! Good eyes.

jonhoo: What kind of maintainer would I be if I didn't actually read the diff in PRs? ;)

marshallpierce: Do you ever sleep, is what I want to know... :P

jonhoo: Hah, I could ask you the same thing! The rate and scope of these PRs is impressive.

marshallpierce: Well, this project is a delightful intersection of two things I'm interested in. It pains me how poorly most stacks handle collecting, aggregating, and otherwise processing performance measurements, and while I could go work on revamping that JVM world that I'm quite familiar with, I'm quite enjoying Rust right now, so here I am. I feel like I have an opportunity while Rust is young to get out ahead of people doing it wrong and help make the easy path also be the correct path. :) I enjoy making things bottom-up with the finest quality ingredients, building or improving components if I can't find ones that meet my needs. That's not practical for day-job software generally, but on my own time, I can polish as much as I want...

jonhoo: Not sure if this is down your alley, but there's recently been a call for help for the Rust concurrency primitives in [crossbeam](https://internals.rust-lang.org/t/crossbeam-request-for-help/4933). You may find that interesting too :)

marshallpierce: Concurrency is near and dear to my heart. Maybe in a few months! :)

Issue #13, opened 2017-02-15
Ensure usage from correct logic will never panic: use `Result` where appropriate, otherwise document correct usage


marshallpierce:
This will make it possible to guarantee panic-free codepaths.

Optionally, we might expose ways to record, read, that can panic but do not require the `Result` syntactic overhead.

jonhoo:
Implementing `AddAssign` already sort-of gives us the latter, but having named methods for this as well would probably also be a good addition.

marshallpierce:
On further consideration, I'm not sure that a blanket "use Result everywhere" policy is actually beneficial. The potential overflow in `value_for`, for instance, isn't necessarily useful to expose with a `Result` since you will only encounter an `Err` if your logic is wrong and you iterate too far. As an example, what about an iterator implementation that is correctly implemented? It will have `Result`-handling code that is never going to hit error cases unless it has incorrect logic that tries to iterate too far.

I think a more useful goal would be something like:

- Use Result where valid logic can still encounter an overflow, etc, error
- In places that will always succeed unless a programmer error is made, carefully document what the result will be (e.g. "if the input cannot be represented in a `u64`, the result will be garbage, but it will not panic"). This means using `wrapping_shl` and friends to make that behavior more explicit.

jonhoo:
Yeah, you're right, that seems like a reasonable plan.

marshallpierce:
Potential panic? https://github.com/jonhoo/hdrsample/pull/34/files#diff-b4aea3e418ccdb71239b96952d9cddb6R1559

jonhoo:
Closing now that #55 has been merged. #38 is still a concern, but that's tracked, well, in #38.

PR #34, opened 2017-03-15
Count saturates

marshallpierce:
I'd intended to do this later but it got in the way of benchmarks for deserialization to panic whenever things overflowed.

There are more overflows to handle (TODOs added) but they aren't inhibiting me so I left them for later. Each one needs a benchmark and tests, so they take a little time. 

Fortunately, this change has no effect whatsoever on performance, at least as far as I could measure. This is a somewhat tricky thing to benchmark because the thing measured is significantly faster than, say, random number generation, so I did the best I could by precalculating numbers to record (or histograms to add/subtract) and then simply iterating across that vec in the hot loop.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -171,12 +171,14 @@ const ORIGINAL_MAX: u64 = 0;
 /// into an integer count. Partial ordering is used for threshholding, also usually in the context
 /// of percentiles.
 pub trait Counter
-    : num::Num + num::ToPrimitive + num::FromPrimitive + Copy + PartialOrd<Self> {
+    : num::Num + num::ToPrimitive + num::FromPrimitive + num::Saturating + num::CheckedSub
+    + num::CheckedAdd + Copy + PartialOrd<Self>{
 }
 
 // auto-implement marker trait
 impl<T> Counter for T
-    where T: num::Num + num::ToPrimitive + num::FromPrimitive + Copy + PartialOrd<T>
+    where T: num::Num + num::ToPrimitive + num::FromPrimitive + num::Saturating + num::CheckedSub
+    + num::CheckedAdd + Copy + PartialOrd<T>
 {
 }
 
@@ -455,13 +457,14 @@ impl<T: Counter> Histogram<T> {
             for i in 0..source.len() {
                 let other_count = source[i];
                 if other_count != T::zero() {
-                    self[i] = self[i] + other_count;
+                    self[i] = self[i].saturating_add(other_count);
                     // TODO unwrapping .to_u64()
-                    observed_other_total_count = observed_other_total_count + other_count.to_u64().unwrap();
+                    observed_other_total_count = observed_other_total_count
+                        .saturating_add(other_count.to_u64().unwrap());
                 }
             }
 
-            self.total_count = self.total_count + observed_other_total_count;
+            self.total_count = self.total_count.saturating_add(observed_other_total_count);
             let mx = source.max();
             if mx > self.max() {
                 self.update_max(mx);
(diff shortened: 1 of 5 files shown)


src/lib.rs
```diff
@@ -552,6 +555,9 @@ impl<T: Counter> Histogram<T> {
             if other_count != T::zero() {
                 let other_value = other.value_for(i);
                 if self.count_at(other_value).unwrap() < other_count {
+                    // TODO Perhaps we should saturating sub here? Or expose some form of
+                    // pluggability so users could choose to error or saturate? Both seem useful.
+                    // It's also sort of inconsistent with overflow, which now saturates.
```
jonhoo: Hmm, yes, it's a good question.. 
One thing to do would be to have a variant in `SubtractionError` (and possibly also `AdditionError`) called `PartialResult` (or something like that), which contains the `Histogram` when using saturating subtract/add. That way the user can either chose to deal with it as an error, or they can match that case and ignore the error.

src/lib.rs
```diff
@@ -777,11 +783,19 @@ impl<T: Counter> Histogram<T> {
     }
 
     fn alter_n(&mut self, value: u64, count: T, add: bool) -> Result<(), ()> {
+        // TODO consider split out addition and subtraction cases; this isn't gaining much by
+        // unifying since we have to test all the cases anyway, and the TODO below is marking a case
+        // that might well be impossible but seems needed because of the (possibly false) symmetry
+        // with addition
```
jonhoo: Good point. I agree.

src/lib.rs
```diff
@@ -798,9 +812,14 @@ impl<T: Counter> Histogram<T> {
...
-                    *c = *c + count;
+                    // after resize, should be no possibility of overflow because this is a new slot
+                    *c = (*c).checked_add(&count).expect("count overflow after resize");
                 } else {
-                    *c = *c - count;
+                    // TODO Not sure this code path can ever be hit: if subtraction requires minuend
+                    // count to exceed subtrahend count for a given value, we shouldn't ever need
+                    // to resize to subtract.
+                    // Anyway, at the very least, we know it shouldn't underflow.
+                    *c = (*c).checked_sub(&count).expect("count underflow after resize");
```
jonhoo: Maybe make this `unreachable!()` then?

src/lib.rs
```diff
@@ -1229,6 +1252,7 @@ impl<T: Counter> Histogram<T> {
     ///
     /// Note that the return value is capped at `u64::max_value()`.
     pub fn median_equivalent(&self, value: u64) -> u64 {
+        // TODO isn't this just saturating?
```
jonhoo: Yup. I guess I just didn't know about `saturating_add` at the time. Happy to have it changed.

src/lib.rs
```diff
@@ -1533,6 +1556,7 @@ impl<T: Counter, F: Counter> PartialEq<Histogram<F>> for Histogram<T>
         if self.min_nz() != other.min_nz() {
             return false;
         }
+        // TODO may panic? Does the above guarantee that the other array is at least as long?
```
jonhoo: Ah, good point. Yes, this has a bug.

jonhoo:
Looks good. I'll leave it open to hear your thoughts on the inline comments above, but happy to merge after that.

src/lib.rs
marshallpierce: Filed as https://github.com/jonhoo/hdrsample/issues/36

src/lib.rs
marshallpierce: It would take more subtraction testing to make me feel confident that this is in fact unreachable, so I'd prefer to wait for #36 for that change

src/lib.rs
marshallpierce: Filed as #37

src/lib.rs
marshallpierce: Can do.

marshallpierce: Actually, it looks like this should never overflow, so I'll use checked_add (and a different PR, just in case that discussion might take us in a different direction from this PR)

src/lib.rs
marshallpierce: Linked from #13.

src/lib.rs
marshallpierce: PR #39.

PR #39, opened 2017-03-15
Simplify overflowing_add in median_equivalent

marshallpierce:
It looks like in fact that logic can (should) never overflow, so use checked_add instead.

I couldn't think of how that could overflow; am I missing some scenario that guided you to implement overflow handling in the first place?


jonhoo:
Can you rebase following the merge of #34?
I believe this was actually a direct port of the HdrHistogram as written at the time. If you're convinced it can't overflow, I believe you.

marshallpierce:
Rebased.

The current logic in the Java version also doesn't check for overflow, so that's encouraging I guess :)

PR #42, opened 2017-03-18
Fix occasional test failures caused by incorrect min value tracking

marshallpierce:
While working on something else I implemented better test random number generation, and sure enough it uncovered a bug. It would cause the minimum value to be set too low with certain patterns of non-zero value counts.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -173,7 +173,7 @@ const ORIGINAL_MAX: u64 = 0;
 /// of percentiles.
 pub trait Counter
     : num::Num + num::ToPrimitive + num::FromPrimitive + num::Saturating + num::CheckedSub
-    + num::CheckedAdd + Copy + PartialOrd<Self>{
+    + num::CheckedAdd + Copy + PartialOrd<Self> {
 }
 
 // auto-implement marker trait
--- a/src/serialization/deserializer.rs
+++ b/src/serialization/deserializer.rs
@@ -240,12 +240,11 @@ pub fn zig_zag_decode(encoded: u64) -> i64 {
 /// We need to perform the same logic in two different decode loops while carrying over a modicum
 /// of state.
 struct DecodeLoopState<T: Counter> {
-    dest_index:usize,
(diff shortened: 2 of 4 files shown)


src/lib.rs
```diff
@@ -173,7 +173,7 @@ const ORIGINAL_MAX: u64 = 0;
 /// of percentiles.
 pub trait Counter
     : num::Num + num::ToPrimitive + num::FromPrimitive + num::Saturating + num::CheckedSub
-    + num::CheckedAdd + Copy + PartialOrd<Self>{
+    + num::CheckedAdd + Copy + PartialOrd<Self> {
```
marshallpierce: I keep on adding and removing `Debug` as needed for printlns. At least now it has a space...

src/serialization/deserializer.rs
```diff
@@ -266,10 +265,12 @@ impl <T: Counter> DecodeLoopState<T> {
...
                 .ok_or(DeserializeError::UnsuitableCounterType)?;
 
-            h.set_count_at_index(self.dest_index, count)
-                .map_err(|_| DeserializeError::EncodedArrayTooLong)?;
+            if count > T::zero() {
+                h.set_count_at_index(self.dest_index, count)
+                    .map_err(|_| DeserializeError::EncodedArrayTooLong)?;
 
-            restat_state.on_nonzero_count(self.dest_index, count);
+                restat_state.on_nonzero_count(self.dest_index, count);
```
marshallpierce: You'd think the `on_nonzero_count` would have been enough of a clue for me when writing the original logic, but apparently not. :/

src/lib.rs
jonhoo: You may want to just install `rustfmt` and have it run on every save. That should avoid these kinds of issues.

marshallpierce: Yeah, that's not a bad idea... On other projects I've set up git hooks, etc; maybe I'll port one of those over to make it easy for others to adopt. Consistency is good.

PR #43, opened 2017-04-06
V2 + Deflate serialization support with flate2's stream wrappers

marshallpierce:
Performance isn't awful but it isn't great either. Compression is slow and there's not a lot we can do about that.

```
test deserialize_large_dense_v2          ... bench:  18,970,418 ns/iter (+/- 944,849)
test deserialize_large_dense_v2_deflate  ... bench:  23,991,379 ns/iter (+/- 1,329,386)
test deserialize_large_sparse_v2         ... bench:  17,710,620 ns/iter (+/- 1,145,763)
test deserialize_large_sparse_v2_deflate ... bench:  20,297,392 ns/iter (+/- 1,568,093)

test serialize_large_dense_v2            ... bench:   7,154,361 ns/iter (+/- 307,599)
test serialize_large_dense_v2_deflate    ... bench:  94,926,041 ns/iter (+/- 3,333,186)
test serialize_large_sparse_v2           ... bench:   6,570,606 ns/iter (+/- 639,558)
test serialize_large_sparse_v2_deflate   ... bench:  63,914,342 ns/iter (+/- 4,191,686)
```

### DIFF
--- a/benches/serialization.rs
+++ b/benches/serialization.rs
@@ -9,107 +9,131 @@ use hdrsample::serialization::*;
 use self::rand::distributions::range::Range;
 use self::rand::distributions::IndependentSample;
 use self::test::Bencher;
-use std::io::Cursor;
+use std::io::{Cursor, Write};
+use std::fmt::Debug;
 
 #[bench]
-fn serialize_tiny_dense(b: &mut Bencher) {
+fn serialize_tiny_dense_v2(b: &mut Bencher) {
     // 256 + 3 * 128 = 640 counts
-    do_serialize_bench(b, 1, 2047, 2, 1.5)
+    do_serialize_bench(b, &mut V2Serializer::new(), 1, 2047, 2, 1.5)
 }
 
 #[bench]
-fn serialize_tiny_sparse(b: &mut Bencher) {
+fn serialize_tiny_sparse_v2(b: &mut Bencher) {
     // 256 + 3 * 128 = 640 counts
-    do_serialize_bench(b, 1, 2047, 2, 0.1)
+    do_serialize_bench(b, &mut V2Serializer::new(), 1, 2047, 2, 0.1)
 }
 
 #[bench]
-fn serialize_small_dense(b: &mut Bencher) {
+fn serialize_small_dense_v2(b: &mut Bencher) {
     // 2048 counts
-    do_serialize_bench(b, 1, 2047, 3, 1.5)
+    do_serialize_bench(b, &mut V2Serializer::new(), 1, 2047, 3, 1.5)
 }
 
 #[bench]
-fn serialize_small_sparse(b: &mut Bencher) {
+fn serialize_small_sparse_v2(b: &mut Bencher) {
     // 2048 counts
-    do_serialize_bench(b, 1, 2047, 3, 0.1)
+    do_serialize_bench(b, &mut V2Serializer::new(), 1, 2047, 3, 0.1)
 }
 
 #[bench]
-fn serialize_medium_dense(b: &mut Bencher) {
+fn serialize_medium_dense_v2(b: &mut Bencher) {
     // 56320 counts
-    do_serialize_bench(b, 1, u64::max_value(), 3, 1.5)
+    do_serialize_bench(b, &mut V2Serializer::new(), 1, u64::max_value(), 3, 1.5)
 }
 
 #[bench]
-fn serialize_medium_sparse(b: &mut Bencher) {
+fn serialize_medium_sparse_v2(b: &mut Bencher) {
     // 56320 counts
-    do_serialize_bench(b, 1, u64::max_value(), 3, 0.1)
+    do_serialize_bench(b, &mut V2Serializer::new(), 1, u64::max_value(), 3, 0.1)
 }
 
 #[bench]
-fn serialize_large_dense(b: &mut Bencher) {
+fn serialize_large_dense_v2(b: &mut Bencher) {
     // 6291456 buckets
-    do_serialize_bench(b, 1, u64::max_value(), 5, 1.5)
+    do_serialize_bench(b, &mut V2Serializer::new(), 1, u64::max_value(), 5, 1.5)
 }
 
 #[bench]
-fn serialize_large_sparse(b: &mut Bencher) {
+fn serialize_large_sparse_v2(b: &mut Bencher) {
     // 6291456 buckets
-    do_serialize_bench(b, 1, u64::max_value(), 5, 0.1)
+    do_serialize_bench(b, &mut V2Serializer::new(), 1, u64::max_value(), 5, 0.1)
 }
 
 #[bench]
-fn deserialize_tiny_dense(b: &mut Bencher) {
+fn serialize_large_dense_v2_deflate(b: &mut Bencher) {
+    // 6291456 buckets
+    do_serialize_bench(b, &mut V2DeflateSerializer::new(), 1, u64::max_value(), 5, 1.5)
+}
+
+#[bench]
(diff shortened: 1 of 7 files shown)


benches/serialization.rs
```diff
@@ -149,3 +173,30 @@ fn do_deserialize_bench(b: &mut Bencher, low: u64, high: u64, digits: u8, fracti
         let _: Histogram<u64> = d.deserialize(&mut cursor).unwrap();
     });
 }
+
+// Maybe someday there will be an obvious right answer for what serialization should look like, at
+// least to the user, but for now we'll only take an easily reversible step towards that. There are
+// still several ways the serializer interfaces could change to achieve better performance, so
+// committing to anything right now would be premature.
+trait TestOnlyHypotheticalSerializerInterface {
```
marshallpierce: This, and its cousin in the private test module, are regrettable, but I still have a handful of TODOs to explore in serialization so I don't want to commit to anything just yet. Plus, I suspect it will not be common that users want to polymorphically choose which serializer they use (if for no other reason than that this feature is so new...).

src/serialization/deserializer.rs
```diff
@@ -57,10 +58,29 @@ impl Deserializer {
...
+            _ => Err(DeserializeError::InvalidCookie)
+        }
+    }
+
+    fn deser_v2_compressed<T: Counter, R: Read>(&mut self, reader: &mut R) -> Result<Histogram<T>, DeserializeError> {
+        let payload_len = reader.read_u32::<BigEndian>()?.to_usize()
+            .ok_or(DeserializeError::UsizeTypeTooSmall)?;
+
+        // TODO reuse deflate buf, or switch to lower-level flate2::Decompress
+        let mut deflate_reader = DeflateDecoder::new(reader.take(payload_len as u64));
```
marshallpierce: allocation :'(

src/serialization/v2_deflate_serializer.rs
```diff
@@ -0,0 +1,86 @@
...
+            .map_err(|e| V2DeflateSerializeError::InternalSerializationError(e))?;
+
+        debug_assert_eq!(self.uncompressed_buf.len(), uncompressed_len);
+
+        self.compressed_buf.write_u32::<BigEndian>(V2_COMPRESSED_COOKIE)?;
+        // placeholder for length
+        self.compressed_buf.write_u32::<BigEndian>(0)?;
+
+        // TODO pluggable compressors? configurable compression levels?
+        // TODO benchmark https://github.com/sile/libflate
```
marshallpierce: So many TODOs! But none of them are errors, just things I want to explore.

src/serialization/serialization.rs
```diff
@@ -86,6 +90,8 @@
 //!     fn serialize<S: Serializer>(&self, serializer: S) -> Result<(), ()> {
 //!         // not optimal to not re-use the vec and serializer, but it'll work
 //!         let mut vec = Vec::new();
+//!         // pick the format you want to use
```
jonhoo: I'd like to see this expanded on a little

src/serialization/deserializer.rs
jonhoo: Is it the allocation of the `DefalteDecoder` you're saddened about? It's a relatively minor allocation, and it's not data-dependent, so I don't know that it'll add significant overhead..?

src/serialization/serialization.rs
```diff
@@ -176,13 +183,19 @@ mod benchmarks;
 mod v2_serializer;
 pub use self::v2_serializer::{V2Serializer, V2SerializeError};
 
+#[path = "v2_deflate_serializer.rs"]
```
jonhoo: You know that Rust, when confronted with `mod foo`, will look for both `foo/mod.rs` *and* `./foo.rs` by default, right? I don't think this `#[path]` should be necessary.

benches/serialization.rs
jonhoo: It'd make me happy if we could at least somehow share this trait and its `impl`s between the benchmarks and the tests though.

src/serialization/v2_deflate_serializer.rs
```diff
@@ -0,0 +1,86 @@
...
+                                           -> Result<usize, V2DeflateSerializeError> {
+        // TODO benchmark serializing in chunks rather than all at once: each uncompressed v2 chunk
+        // could be compressed and written to the compressed buf, possibly using an approach like
+        // that of https://github.com/jonhoo/hdrsample/issues/32#issuecomment-287583055.
+        // This would reduce the overall buffer size needed for plain v2 serialization, and be
+        // more cache friendly.
+
+        self.uncompressed_buf.clear();
+        self.compressed_buf.clear();
+        // TODO serialize directly into uncompressed_buf without the buffering inside v2_serializer
```
jonhoo: Ahh, yeah, that's unfortunate :(

src/serialization/v2_deflate_serializer.rs
```diff
@@ -0,0 +1,86 @@
...
+        self.compressed_buf.write_u32::<BigEndian>(V2_COMPRESSED_COOKIE)?;
+        // placeholder for length
+        self.compressed_buf.write_u32::<BigEndian>(0)?;
+
+        // TODO pluggable compressors? configurable compression levels?
+        // TODO benchmark https://github.com/sile/libflate
+
+        {
+            // TODO reuse deflate buf, or switch to lower-level flate2::Compress
+            let mut compressor = DeflateEncoder::new(&mut self.compressed_buf, Compression::Default);
```
jonhoo: I wonder if we could speed up first-time serialization pretty substantially by resizing `self.compressed_buf` here to some fraction of `self.uncompressed_buf.len()`. This would avoid doing multiple `Vec` resizes during the first compression.

benches/serialization.rs
marshallpierce: Yeah, I would like that as well but I couldn't think of a way to do it without making it `pub`. Thoughts?

src/serialization/deserializer.rs
marshallpierce: Yeah, but mostly for intellectual vanity reasons (and a little bit because I want to keep the way to no-heap open) I want serialization and deserialization to be allocation-free. :)

src/serialization/v2_deflate_serializer.rs
marshallpierce: Sure, why not. That would make the pessimal case of single-use serializers less bad!

src/serialization/serialization.rs
marshallpierce: You'd think that, and yet without this...

```
error: cannot declare a new module at this location
   --> src/serialization/serialization.rs:186:1
    |
186 | mod v2_deflate_serializer;
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
    = note: for more information, see issue #37872 <https://github.com/rust-lang/rust/issues/37872>
```

benches/serialization.rs
jonhoo: I *think* you may be able to (ab)use the `#[path]` syntax to include a file with this trait + `impl`s as a module in both...

src/serialization/serialization.rs
jonhoo: Hmm.. Reading issue https://github.com/rust-lang/rust/issues/37872, it seems like this is caused by us using `#[path]` elsewhere, which sort "taints" the path such that everything needs `#[path]` annotations. Weird... And unfortunate..

jonhoo:
LGTM!

PR #46, opened 2017-04-06
Port record_correct benchmark from Criterion.

### LINKED ISSUE #30: Port Criterion benchmarks to use built-in benchmarking
It would be nice to have only one benchmark tool in use, and Criterion doesn't seem to have much momentum.

marshallpierce:
Removed other Criterion benchmark and busy loop as they already
have representation in the existing benchmarks. Fixes #30.


jonhoo:
I do feel a little sad losing the warm-up time that Criterion provides, but it's still the right change to make.

marshallpierce:
Yes, that's one of the many things I wish Rust's built-in tool would do. I miss JMH :/

PR #48, opened 2017-04-12
Split alter_n into dedicated add and subtract logic.

marshallpierce:
On master:

```
test add_precalc_random_value_1_count_different_precision_u64   ... bench:  10,614,189 ns/iter (+/- 252,423)
test add_precalc_random_value_1_count_same_dimensions_u64       ... bench:  46,938,026 ns/iter (+/- 843,737)
test add_precalc_random_value_max_count_different_precision_u64 ... bench:  10,595,457 ns/iter (+/- 214,227)
test add_precalc_random_value_max_count_same_dimensions_u64     ... bench:  46,969,248 ns/iter (+/- 670,218)
test record_correct_precalc_random_values_with_1_count_u64      ... bench:  30,194,131 ns/iter (+/- 749,564)
test record_precalc_random_values_with_1_count_u64              ... bench:   5,783,699 ns/iter (+/- 88,471)
test record_precalc_random_values_with_max_count_u64            ... bench:   5,769,106 ns/iter (+/- 162,864)
test record_random_values_with_1_count_u64                      ... bench:   7,911,239 ns/iter (+/- 185,565)
test subtract_precalc_random_value_1_count_same_dimensions_u64  ... bench:  53,987,588 ns/iter (+/- 920,352)
```

Now:

```
test add_precalc_random_value_1_count_different_precision_u64   ... bench:  10,550,182 ns/iter (+/- 456,035)
test add_precalc_random_value_1_count_same_dimensions_u64       ... bench:  46,924,786 ns/iter (+/- 1,495,721)
test add_precalc_random_value_max_count_different_precision_u64 ... bench:  10,522,688 ns/iter (+/- 322,702)
test add_precalc_random_value_max_count_same_dimensions_u64     ... bench:  46,970,544 ns/iter (+/- 1,637,041)
test record_correct_precalc_random_values_with_1_count_u64      ... bench:  26,863,186 ns/iter (+/- 408,375)
test record_precalc_random_values_with_1_count_u64              ... bench:   5,349,910 ns/iter (+/- 140,676)
test record_precalc_random_values_with_max_count_u64            ... bench:   5,350,301 ns/iter (+/- 116,121)
test record_random_values_with_1_count_u64                      ... bench:   7,252,872 ns/iter (+/- 200,447)
test subtract_precalc_random_value_1_count_same_dimensions_u64  ... bench:  62,547,610 ns/iter (+/- 1,258,993)
```

[...]

PR #49, opened 2017-04-12
Only support u8-u64 as counter types. Fixes #29.

### LINKED ISSUE #29: Restrict counts to only u8-u64
We don't seem to be gaining much by allowing, say, `f64`, and it leads to some questionable things like:

```
total_count = total_count + count.to_u64().unwrap();
```

Meaningfully handling the case where that *doesn't* work seems iffy.

marshallpierce:
This will have a minor conflict with #48; I'll fix up whichever one doesn't get merged first.


jonhoo:
LGTM. Thanks!

PR #52, opened 2017-08-19
Use saturating math to avoid a couple TODOs.

marshallpierce:
Also, use u64 when summing counts for better accuracy.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1216,11 +1216,11 @@ impl<T: Counter> Histogram<T> {
         }
 
         let target_index = cmp::min(self.index_for(value).ok_or(())?, self.last());
-        // TODO overflow
         // TODO panic on bad index
-        let total_to_current_index =
-            (0..(target_index + 1)).map(|i| self[i]).fold(T::zero(), |t, v| t + v);
-        Ok(100.0 * total_to_current_index.as_f64() / self.total_count as f64)
+        let total_to_current_index = (0..(target_index + 1))
+            .map(|i| self[i].as_u64())
+            .fold(0_u64, |t, v| t.saturating_add(v));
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -1216,11 +1216,11 @@ impl<T: Counter> Histogram<T> {
...
         let target_index = cmp::min(self.index_for(value).ok_or(())?, self.last());
-        // TODO overflow
         // TODO panic on bad index
-        let total_to_current_index =
-            (0..(target_index + 1)).map(|i| self[i]).fold(T::zero(), |t, v| t + v);
-        Ok(100.0 * total_to_current_index.as_f64() / self.total_count as f64)
+        let total_to_current_index = (0..(target_index + 1))
+            .map(|i| self[i].as_u64())
+            .fold(0_u64, |t, v| t.saturating_add(v));
+        Ok(100.0 * total_to_current_index as f64 / self.total_count as f64)
```
jonhoo: Hmm.. Do we also make sure we use saturating add on `self.total_count`?

marshallpierce: Yep, looks like we've got that in place:

- in `add` https://github.com/jonhoo/hdrsample/blob/master/src/lib.rs#L528
- in `subtract` (underflow) https://github.com/jonhoo/hdrsample/blob/master/src/lib.rs#L646
- in `record_n` https://github.com/jonhoo/hdrsample/blob/master/src/lib.rs#L884
in `restat` https://github.com/jonhoo/hdrsample/blob/master/src/lib.rs#L1539

Looks like that has pretty good test coverage too: https://github.com/jonhoo/hdrsample/blob/master/tests/histogram.rs#L831 onwards

PR #54, opened 2017-08-19
Simplify usize calculation

marshallpierce:
Includes #52 and #53, so peruse those first.

`count_at` and a couple of others don't really need a `Result` return type, it turns out, so this is a nice ergo win. The Java implementation clamps oversized values to the max representable value, so we can too!

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1227,6 +1227,7 @@ impl<T: Counter> Histogram<T> {
 
         let target_index = self.index_for_or_last(value);
         // TODO panic on bad index
+        // TODO use RangeInclusive when it's stable to avoid checked_add
         let total_to_current_index = (0..target_index.checked_add(1).expect("usize overflow"))
             .map(|i| self[i].as_u64())
             .fold(0_u64, |t, v| t.saturating_add(v));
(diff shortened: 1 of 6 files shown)


src/lib.rs
```diff
@@ -1207,20 +1215,22 @@ impl<T: Counter> Histogram<T> {
...
+    /// inaccurate results.
+    pub fn percentile_below(&self, value: u64) -> f64 {
         if self.total_count == 0 {
-            return Ok(100.0);
+            return 100.0;
         }
 
-        let target_index = cmp::min(self.index_for(value).ok_or(())?, self.last());
-        // TODO overflow
+        let target_index =self.index_for_or_last(value);
```
jonhoo: Missing space

jonhoo:
Mmm, this is a good change.

jonhoo:
Can you rebase onto master?

marshallpierce:
Rebased.

src/lib.rs
```diff
@@ -1207,20 +1215,22 @@ impl<T: Counter> Histogram<T> {
...
         if self.total_count == 0 {
-            return Ok(100.0);
+            return 100.0;
         }
 
-        let target_index = cmp::min(self.index_for(value).ok_or(())?, self.last());
+        let target_index = self.index_for_or_last(value);
         // TODO panic on bad index
-        let total_to_current_index = (0..(target_index + 1))
+        let total_to_current_index = (0..target_index.checked_add(1).expect("usize overflow"))
```
jonhoo: Ugh, this is unfortunate. What we really want is [`RangeInclusive`](https://doc.rust-lang.org/std/ops/struct.RangeInclusive.html), but it's currently nightly-only. Perhaps leave a TODO?

marshallpierce: Definitely. I've wanted this more than a few times. :(

PR #56, opened 2017-09-16
Miscellaneous changes

danburkert:
Included are a few small changes.  See the individual commit messages for more details.  I'm happy to pull out / reorder / squash them if deemed necessary.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,3 +7,7 @@ cache: cargo
 matrix:
   allow_failures:
     - rust: nightly
+
+script:
+  - cargo test --verbose
+  - cargo test --verbose --no-default-features --no-run
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1396,6 +1396,7 @@ impl<T: Counter> Histogram<T> {
     }
 
     /// Returns an error if the index doesn't exist.
+    #[cfg(feature = "serialization")]
     fn set_count_at_index(&mut self, index: usize, count: T) -> Result<(), ()> {
         let r = self.counts.get_mut(index).ok_or(())?;
         *r = count;
@@ -1687,5 +1688,6 @@ impl<T: Counter, F: Counter> PartialEq<Histogram<F>> for Histogram<T>
 #[cfg(test)]
 mod tests;
 
+#[cfg(feature = "serialization")]
 #[path = "serialization/serialization.rs"]
 pub mod serialization;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -181,7 +181,7 @@
 // Enable feature(test) is enabled so that we can have benchmarks of private code
 #![cfg_attr(all(test, feature = "bench_private"), feature(test))]
 
-extern crate num;
+extern crate num_traits as num;
 
 use std::borrow::Borrow;
 use std::cmp;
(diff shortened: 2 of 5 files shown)


.travis.yml
```diff
@@ -7,3 +7,8 @@ cache: cargo
 matrix:
   allow_failures:
     - rust: nightly
+
+script:
+  - cargo build --verbose
+  - cargo test --verbose
+  - cargo build --verbose --no-default-features
```
jonhoo: This is also missing
```yaml
  - cargo test --verbose --no-default-features
```
Maybe I missed it, but aren't you missing `cfg(serialization)` attributes on the serialization test cases too?

Also, what is the advantage of running `cargo build` and `cargo test` as separate tests? I suppose in case the code has `cfg(test)` parts in it?

jonhoo:
The `num-traits` change and `serialization` feature seem like good ideas to me. `impl Counter for usize` I want to double-check with @marshallpierce. See also inline comment about `.travis.yml`.

.travis.yml
danburkert: build + test being separate doesn't really have an advantage.  As far as testing with `--no-default-features`, I didn't think it was worthwhile to run the test suite again, and it requires adding another `cfg`, but that's easy enough so I'll go ahead and make that change.

danburkert: Updated.

.travis.yml
```diff
@@ -7,3 +7,7 @@ cache: cargo
 matrix:
   allow_failures:
     - rust: nightly
+
+script:
+  - cargo test --verbose
+  - cargo test --verbose --no-default-features
```
jonhoo: I guess `cargo test --no-default-features` will (at least in this case) run strictly fewer tests. However, it'll be useful to check that the tests at least still compile when serialization is disabled, so I'd use:
```yaml
cargo test --verbose --no-default-features --no-run
```

danburkert: Good idea.

src/lib.rs
```diff
@@ -239,6 +239,13 @@ impl Counter for u64 {
     fn as_u64(&self) -> u64 { *self }
 }
 
+impl Counter for usize {
+    #[inline]
+    fn as_f64(&self) -> f64 { *self as f64 }
+    #[inline]
+    fn as_u64(&self) -> u64 { *self as u64 }
+}
```
marshallpierce: This is probably safe to add, though I haven't thought through it carefully yet. However, I'm not clear why this is desirable. Can you maybe explain your use-case a little more? Typically people would choose a counter type based on the sample rate and the duration that a single histogram would be accumulating data. Why is it desirable to have the maximum count per bucket change when the architecture changes?

danburkert: Initially I added `usize` because the `tacho` library uses a `usize` counter, and I'm updating it from hdrsample 3.x to 4.x.  That being said, I think your question is valid, and I didn't quite think this through.  It's likely that `tacho` should be using a fixed size value.  I'm going to strip this commit from the PR.

marshallpierce: Ah, makes sense. Glad to see this data structure being used for monitoring stuff!

danburkert:
I've removed the commit adding a `usize` counter.

jonhoo:
Great, thank you!

PR #57, opened 2017-09-17
Release 5.0

jonhoo:
Now that we have a bunch of new changes and improvements, I suggest we cut a new release and publish to crates.io. I also think it's high time you're listed as an author @marshallpierce! I can add you to the crates.io as a crate owner too if you're interested?

/cc @danburkert

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,11 +1,14 @@
 [package]
 name = "hdrsample"
-version = "4.0.0"
+version = "5.0.0"
 
 description = "A port of HdrHistogram to Rust"
 readme = "README.md"
 
-authors = ["Jon Gjengset <jon@thesquareplanet.com>"]
+authors = [
+	"Jon Gjengset <jon@thesquareplanet.com>",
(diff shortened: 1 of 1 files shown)


marshallpierce:
Hooray, I've graduated ;) I'm happy to be a crates.io owner too.

I do have one vaguely breaking change waiting in the wings about percentile calculations (see enormous chat log in gitter hdrhistogram room). I don't have the branch ready immediately, though, so it would mean waiting a few days for me to finish assembling it. If there isn't a pressing need for a release immediately, I'd like to include that change in the next major release train. Thoughts?

jonhoo:
Sure, then let's wait for that. I was mostly thinking of @danburkert who seems like he might want his new changes and sync with the latest `hdrsample` (which is beyond 4.0). But I'm okay with waiting.

danburkert:
I'm not in a hurry :)

marshallpierce:
OK, great. I'm out of town this weekend but I should be able to get it PRable by Tuesday or so.

jonhoo:
Merged and published!

PR #58, opened 2017-09-20
Quantile/percentile calculation

marshallpierce:
This is the result of much discussion in the [gitter](https://gitter.im/HdrHistogram/HdrHistogram) room.

The important change is that `value_for_quantile` (which was `value_for_percentile`) now calculates the count for a particular quantile with different rounding behavior.

It turns out that the [original issue](https://github.com/HdrHistogram/HdrHistogram_py/pull/14) was basically a (mostly) unavoidable consequence of IEEE754 floating point precision loss, coupled with the not-quite-optimal way we were calculating percentile/quantile counts. The current arithmetic around quantiles is what I think is the most mathematically defensible way to do it.

I got all aggro on nixing percentiles as a concept because they offer no value beyond aesthetics over quantiles, but involve extra FP multiplications and divisions and therefore lose precision. I left a couple `percentile` flavored helpers that seemed relatively inoffensive. I could be talked back from my annihilation of `percentile` as a concept, but I figure the sooner we start training users to use more precise concepts (and avoiding FP issues), the better. (Or maybe we should rip the bandaid off all at once and remove the last few percentile helpers?)

Also, I left the commit history for two reasons: one, given the complex history of this issue, I thought it might be helpful to future maintainers figuring out why we did this when gitter is long gone, and two, fixing the conflicts for the earlier commits was a real pain. ;) If it's too agonizingly ugly for you, I'll go in with a git chainsaw and clean it up.


jonhoo:
For posterity, a more direct link to the discussion is: https://gitter.im/HdrHistogram/HdrHistogram?at=599f36fbee5c9a4c5ff6f438

README.md
```diff
@@ -101,7 +101,7 @@ use hdrsample::Histogram;
 let hist = Histogram::<u64>::new(2).unwrap();
 // ...
 println!("# of samples: {}", hist.count());
-println!("99.9'th percentile: {}", hist.value_at_percentile(99.9));
+println!("99.9'th percentile: {}", hist.value_at_quantile(0.999));
```
jonhoo: I assume this should now say "quantile" in the printed string too?

jonhoo:
This looks good to me, and I have no problem with the move to quantile. I do think leaving the helpers in is a good thing for now though. Thanks to [RFC 1270](https://github.com/rust-lang/rfcs/blob/master/text/1270-deprecation.md), we could mark the percentile functions as deprecated, but I'm not sure that's necessary.

As for leaving the git history, I am all in favor of that. Especially for as gritty a problem as this one!

PR #63, opened 2017-10-10
Fix compression format: needs to have the zlib header around deflate.

### LINKED ISSUE #62: Percentile iterator example for .hgrm output (for plotting)
Hi,

is there an example, how to use the percentile iterator to create a text output that can be plotted like on the hdrhistogram.org Web site's example?

Jan

marshallpierce:
Also add an example cli tool to make it easy to construct serialized
histograms of various flavors. Seems like it might come in handy for
someone beyond me.

Fixes #62.


jonhoo:
Hmm, `tests/serialization.rs` fails to compile: https://travis-ci.org/jonhoo/hdrsample/jobs/285870400#L810

jonhoo:
Also, for the command-line tool, it'd be nice to also have an argument to turn on auto-resize.
Apart from that LGTM.

algermissen:
Thanks for doing this so quickly!

I just tried the branch version:
- Added impl Counter for i32 locally
- produced binary with

    let mut buf = File::create("foo2.hist").expect("ok io");
    let nbytes = V2DeflateSerializer::new()
      .serialize(&hist, &mut buf).expect("ok");
    buf.sync_all().expect("OK");

- The binary is attached
- The base64 version is

    HISTFAAAAFl4AS2KOw5AUBREr+OJiOIVopOIiFKh0ShQ2JfCDizD3sQS3PeZTOYkM9NcdyUimwQRmfic33P9QvHgZBCMmhQLGYUiV1pKWhZf73Qw6ESt74ORnokfyTUHhg==

[foo2.hist.zip](https://github.com/jonhoo/hdrsample/files/1370786/foo2.hist.zip)

Result:
- base64 works(!) in https://hdrhistogram.github.io/HdrHistogramJSDemo/decoding-demo.html

marshallpierce:
Ah, there's a build with no default features. I'll add `cfg` as needed. Auto resize sounds good too.

@algermissen Glad it's working! Adding `Counter` for `i32` shouldn't really be doing anything important, though. Negative values aren't supported anyway. I tried it with just plain old `u64` and interop worked with the JS demo and some throwaway tools I made with the Java impl.

One other thing is that this is a backwards-incompatible change *if* people have caches of old histograms serialized with the plain deflate compression. If that is an issue for people we can resuscitate the old form to let them transcode the old form into the new form.

tests/serialization.rs
```diff
@@ -1,57 +1,61 @@
-extern crate hdrsample;
+#[cfg(feature = "serialization")]
+#[cfg(test)]
```
jonhoo: You can use `all(feature = "serialization", test)` here I believe

jonhoo:
Ah, yes, that came as a result of #56.

For the example binary, we should use [`required-features`](http://doc.crates.io/manifest.html#the-required-features-field-optional) instead of this nasty whole-module `cfg` gate.

As for backwards compatibility, I think we should make the change, and then provide a fix for those in need if complaints should arise.

jonhoo:
@marshallpierce if you think we might want to add other things, perhaps this should become a `bin`, and not just an `example`?

marshallpierce:
Not a bad idea, but I think I'd like to incubate things a bit as an example before we ship it as something semi-official.

PR #65, opened 2017-10-11
Some reorganization and cleanup

soro:
Just moved some things into a core module and cleaned up a tiny bit.


jonhoo:
Thanks! I think this is overall a good change, though I think I'd like at least `error` to be a submodule directly under `src/`. That seems to be the convention among many Rust crates.

soro:
They are reexported at the root level, so a user of the library will be able to import them as hdrsample::someerror. Among my local changes I've also reexported them from core, so the imports are less verbose within the library.
Does that seem fine to you?

jonhoo:
I think the re-export is important, but I'd still like errors to be in the relative standard location of `::error` or `::errors`. I'm fine with `Countner` being in `core`.

soro:
Since I have some time to quickly fix this - you would want them to be externally available at the top level under errors:: or as part of the root import? Currently they are available both under errors:: as well as at the top level.

jonhoo:
Exactly like that! Thanks.

jonhoo:
Also, cc @marshallpierce

Issue #38, opened 2017-03-15
Address remaining logic that can overflow


marshallpierce:
Several TODOs were added in #34 pointing out places that can overflow. There may be more; that was simply what I found by searching for " + ".

PR #68, opened 2017-10-24
Rework iteration to avoid overflow

### LINKED ISSUE #47: Iteration stops too early when total_count has saturated
If `total_count` has reached `u64::max_value()`, iteration should continue even if the total count seen so far during iteration has reached that value. 

I added some tests around that [here](https://github.com/marshallpierce/hdrsample/tree/iter-recorded-total-count).

marshallpierce:
Incorporating a gaggle of smaller fixes I came upon while working on this:

- Iteration now stops when the index with the max value is reached rather than going by total count. This avoids stopping too early when total count saturates (fixes #47, and a few parts of #38). Count since last iteration is also started from 0 at each iteration point, making it less prone to overflow.
- Linear iteration was stopping one iteration too soon in some cases due to a subtle off-by-one.
- Quantile iteration now skips straight to quantile 1.0 if it reaches the index with the max value, just like the Java impl. See long-winded comments, and even more long-winded tests.
- Rather than having pickers provide accessors for metadata about what was just picked, have `pick()` return a metadata struct with optional fields so that the timeline is clear and pickers don't need mostly useless fields to hang on to data until it can be requested 1 function call later.
- Ported iteration tests from the old rust implementation.

### DIFF
--- a/src/iterators/mod.rs
+++ b/src/iterators/mod.rs
@@ -25,14 +25,14 @@ pub struct PickMetadata {
 
     /// Supply the value iterated to in the last `pick()`, if the picker can supply a more useful
     /// value than the largest value represented by the bucket.
-    value_iterated_to: Option<u64>
+    value_iterated_to: Option<u64>,
 }
 
 impl PickMetadata {
     fn new(quantile_iterated_to: Option<f64>, value_iterated_to: Option<u64>) -> PickMetadata {
         PickMetadata {
             quantile_iterated_to,
-            value_iterated_to
+            value_iterated_to,
         }
     }
 }
@@ -45,7 +45,12 @@ pub trait PickyIterator<T: Counter> {
     ///
     /// This will be called with the same index until it returns `None`. This enables modes of
     /// iteration that pick different values represented by the same bucket, for instance.
-    fn pick(&mut self, index: usize, total_count_to_index: u64, count_at_index: T) -> Option<PickMetadata>;
+    fn pick(
+        &mut self,
+        index: usize,
+        total_count_to_index: u64,
+        count_at_index: T,
+    ) -> Option<PickMetadata>;
 
     /// Should we keep iterating even though the last index with non-zero count has already been
     /// picked at least once?
@@ -83,19 +88,24 @@ pub struct IterationValue<T: Counter> {
     quantile: f64,
     quantile_iterated_to: f64,
     count_at_value: T,
-    count_since_last_iteration: u64
+    count_since_last_iteration: u64,
 }
 
 impl<T: Counter> IterationValue<T> {
     /// Create a new IterationValue.
-    pub fn new(value_iterated_to: u64, quantile: f64, quantile_iterated_to: f64, count_at_value: T,
-               count_since_last_iteration: u64) -> IterationValue<T> {
+    pub fn new(
+        value_iterated_to: u64,
+        quantile: f64,
+        quantile_iterated_to: f64,
+        count_at_value: T,
+        count_since_last_iteration: u64,
+    ) -> IterationValue<T> {
         IterationValue {
             value_iterated_to,
             quantile,
             quantile_iterated_to,
             count_at_value,
-            count_since_last_iteration
+            count_since_last_iteration,
         }
     }
 
@@ -113,11 +123,15 @@ impl<T: Counter> IterationValue<T> {
     }
 
     /// Quantile of recorded values that are at or below the current bucket.
-    pub fn quantile(&self) -> f64 { self.quantile }
+    pub fn quantile(&self) -> f64 {
+        self.quantile
+    }
 
     /// Quantile iterated to, which may be different than `quantile()` when an iterator provides
     /// information about the specific quantile it's iterating to.
-    pub fn quantile_iterated_to(&self) -> f64 { self.quantile_iterated_to }
+    pub fn quantile_iterated_to(&self) -> f64 {
+        self.quantile_iterated_to
+    }
 
     /// Recorded count for values equivalent to `value`
     pub fn count_at_value(&self) -> T {
@@ -148,8 +162,9 @@ impl<'a, T: Counter, P: PickyIterator<T>> HistogramIterator<'a, T, P> {
 }
 
 impl<'a, T: 'a, P> Iterator for HistogramIterator<'a, T, P>
-    where T: Counter,
-          P: PickyIterator<T>
+where
+    T: Counter,
+    P: PickyIterator<T>,
 {
     type Item = IterationValue<T>;
     fn next(&mut self) -> Option<Self::Item> {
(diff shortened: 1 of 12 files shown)


src/iterators/mod.rs
```diff
@@ -65,34 +88,35 @@ pub struct IterationValue<T: Counter> {
...
             count_since_last_iteration
         }
     }
 
-    /// The lowest value stored in the current histogram bin
-    pub fn value(&self) -> u64 {
-        self.value
+    /// The value iterated to. Some iterators provide a specific value inside the bucket, while
+    /// others just use the highest value in the bucket.
+    pub fn value_iterated_to(&self) -> u64 {
```
marshallpierce: Maybe this rename-for-clarity is not worth the compatibility concern?

src/iterators/mod.rs
```diff
@@ -155,50 +170,53 @@ impl<'a, T: 'a, P> Iterator for HistogramIterator<'a, T, P>
...
+                            self.count_since_last_iteration.saturating_add(self.count_at_index.as_u64());
 
                     // make sure we don't add this index again
                     self.fresh = false;
                 }
             }
 
             // figure out if picker thinks we should yield this value
-            if self.picker.pick(self.current_index, self.total_count_to_index) {
-                let val = self.current();
```
marshallpierce: I felt it was easier to reason at-a-glance about when the iterator's fields were used with this function inlined, since it matters that you read things like `count_since_last_iteration` *before* resetting a few lines below.

src/iterators/quantile.rs
```diff
@@ -1,13 +1,13 @@
...
 use Histogram;
-use iterators::{HistogramIterator, PickyIterator};
+use iterators::{HistogramIterator, PickyIterator, PickMetadata};
 
 /// An iterator that will yield at quantile steps through the histogram's value range.
 pub struct Iter<'a, T: 'a + Counter> {
     hist: &'a Histogram<T>,
     ticks_per_half_distance: u32,
     quantile_to_iterate_to: f64,
-    quantile_just_picked: f64
```
marshallpierce: this is what returning `Option<PickMetadata>` lets us avoid

src/iterators/quantile.rs
```diff
@@ -21,33 +21,84 @@ impl<'a, T: 'a + Counter> Iter<'a, T> {
...
 
         // This calculation, combined with the `quantile * count` in `value_at_quantile`, tends
         // to produce a count_at_quantile that is 1 ulp wrong. That's just the way IEEE754 works.
         let current_quantile = running_total as f64 / self.hist.count() as f64;
         if current_quantile < self.quantile_to_iterate_to {
-            return false;
+            return None;
+        }
+
+        // Because there are effectively two quantiles in play (the quantile of the value for the
```
marshallpierce: comments per LoC might be getting a little out of hand in this struct, but I like to beat historically confusing logic into submission with overwhelming documentation

tests/data_access.rs
```diff
@@ -542,365 +539,3 @@ fn total_count_exceeds_bucket_type() {
 
     assert_eq!(400, h.count());
 }
-
-#[test]
```
marshallpierce: quantile tests moved to their own file

src/iterators/mod.rs
```diff
@@ -155,50 +170,53 @@ impl<'a, T: 'a, P> Iterator for HistogramIterator<'a, T, P>
...
-
-                // note that we *don't* increment self.current_index here. the picker will be
-                // exposed to the same value again after yielding. not sure why this is the
-                // behavior we want, but it's what the original Java implementation dictates.
-
-                // TODO count starting at 0 each time we emit a value to be less prone to overflow
-                self.prev_total_count = self.total_count_to_index;
+            if let Some(metadata) = self.picker.pick(self.current_index, self.total_count_to_index, self.count_at_index) {
```
jonhoo: I think `rustfmt` might complain about this line.

marshallpierce:
@algermissen here's how the quantile iteration example looks now:

```
jot 10000 0 10000000000 | cg run --example cli -- serialize -c | cg run --example cli -- iter-quantiles -t 5
```

```
       Value          QuantileValue      QuantileIteration TotalCount 1/(1-Quantile)

           0 0.00010000000000000000 0.00000000000000000000          1           1.00
   999292927 0.10000000000000000555 0.10000000000000000555       1000           1.11
  1999634431 0.20000000000000001110 0.20000000000000001110       2000           1.25
  3001024511 0.30009999999999997788 0.30000000000000004441       3001           1.43
...
  9999220735 0.99990000000000001101 0.99985351562500002220       9999        6826.67
  9999220735 0.99990000000000001101 0.99987792968750000000       9999        8192.00
  9999220735 0.99990000000000001101 0.99989013671875004441       9999        9102.22
 10007609343 1.00000000000000000000 0.99990234375000008882      10000       10240.00
 10007609343 1.00000000000000000000 1.00000000000000000000      10000              ∞
#[Mean       = 5000000916.51, StdDeviation   = 2887040392.98]
#[Max        =  10007609343, Total count    =        10000]
#[Buckets    =           54, SubBuckets     =        56320]
```

The `1/(1-quantile)` column is now using the quantile iterated to, not the quantile of the value of the current bucket, so there's only one row with ∞. Also, quantile iteration now skips all intermediate steps to 1.0 once it reaches the last bucket. (Small proviso on that: if total count saturates, it doesn't do the skipping thing. Could be fixed with more fiddling, but this PR was getting complex enough, and probably nobody will ever complain about that...)

jonhoo:
This is *great*! I also think the long comment is useful (not to mention the many extra tests).

See my one comment about `rustfmt`, but apart from that, lgtm.

src/iterators/mod.rs
jonhoo: I think it's fine.

src/iterators/quantile.rs
jonhoo: I have no objection to this comment. Seems well-placed and useful.

src/iterators/mod.rs
marshallpierce: Good point. I haven't been `rustfmt`-ing but I'm happy to adopt it. I don't even really care if its formatting is suboptimal for some aesthetic; consistency is good. I'll format the iterator files.

algermissen:
@marshallpierce Thanks a lot, the iteration logic looks definitely fine now. The double line at end is gone for my cases.

One nit: The CLI output adds a column which breaks the format named 'hgrm' by the original hdrhistogram and CLI output cannot be loaded by the original plot HTML pages. No big deal, just wanted to give that feedback.

marshallpierce:
@algermissen yeah, I'm a little conflicted on that.

Anti-extra-column:
- It's lame to break other tools by tweaking formats.

Pro-anti-column:
- The extended format is a little more comprehensible since it shows why it's possible to have a quantile change without the value changing
- We shouldn't be parsing a fragile text format for machine processing; we should use tools like https://hdrhistogram.github.io/HdrHistogramJSDemo/plotFiles.html that operate on the serialized histogram

Overall I'm weakly in favor of keeping the output the way it is (human oriented) but I could be persuaded.

algermissen:
@marshallpierce I agree with you - I am only building a proof of concept throw-away tool, so I am looking for the least amount of work. Otherwise I'd never use the text format.

As I am fine with copy/pasting your loop and removing the col by hand, no need for me to pursuade you in dropping the col :-)

Keep it as it is - it's better - my 2ct

Issue #59, opened 2017-09-25
Add histogram log format support


marshallpierce:
Generally useful, but also when we develop a corpus of sample serialized histograms with pre-calculated metadata to test an implementation against, it would be good to do the same for the log format -- which means we need the log format first.

marshallpierce:
Fixed in #70.

PR #70, opened 2017-10-27
Interval log support

marshallpierce:
Mostly done; figured I'd get feedback on what I had so far.

The log format has some issues:
- No way to describe what max value scaling factor was used, so you pretty much just have to know what it was if you care about the max value (which you probably shouldn't -- IMO, we should build better tools for displaying that sort of stuff rather than jamming in a semi-human-readable thing, like perhaps one or more metadata modes for our CLI tool to display min, max, etc)
- Some comments are magical metadata comments. It would be better if metadata had a non-comment syntax.
- Metadata can technically be interleaved with interval log lines, but it's unclear if this has any actual meaning
- Tag support is pretty clearly bolted-on, but it's simple enough to parse fortunately.

### Reading logs
This exposes a somewhat low level interface: it doesn't keep track of an observed `BaseTime` and add it to interval histograms' timestamps, for instance. Instead, it exposes an `Iterator` of the useful log lines, from which it would be simple to apply your own preference for dealing with timestamps, skipping logs you don't care about, etc. I decided to try using `nom` for parsing, and I think it came out pretty well.

### Writing logs
Pretty straightforward. I chose to make a `Tag` abstraction so that we could guarantee the lack of tag-forbidden characters in a `str` with the type system. I also split writing duties into two separate types to prevent users from writing lines I think should only exist in the preamble (`StartTime` and `BaseTime`) once they've started writing interval histograms. (Helper methods for those things aren't there yet but the structure is in place.)

### To do
- [x] Figure out what to do `StartTime` and `BaseTime` lines. There's no support for writing them yet mainly because I can't figure out what their semantics are in a clear enough way to document it in a way that isn't  terribly confusing. They seem to sort of clash with each other.
- [X] Benchmarks.
- [X] Builder
- [X] Duration for interval duration
- [X] Duration for interval timestamp
- [X] SystemTime for writing StartTime/BaseTime
- [x] Better type for reading StartTime/BaseTime
- [x] Comment filtering to avoid breaking log format


jonhoo:
I'm confused. Why do we now have our own [`Serialize` trait](https://github.com/jonhoo/hdrsample/pull/70/files#diff-1e8c5d44a44f70e9f67a702dfe981c31R216)?

Will try to look over this at some point in not too long, but flying to a conference in Shanghai tomorrow morning, so may take a little while :)

marshallpierce:
The `Serializer` trait is to abstract over the different serializers (V2, V2 + Deflate) so that they can both be dropped into the log writer as the user prefers.

No big rush on reviewing it; I need some time to ponder the above timestamp issues anyway.

marshallpierce:
I wrote up my interpretation of how the Java impl treats StartTime and BaseTime.

I now have this wishlist for interval log v2:

- Legend line is just a plain comment
- Headers with a defined semantic like StartTime should not be comments, but should have their own syntax
- Users should be able to add their own metadata to the log in a sufficiently structured way that a parser could sanely spit it back to them. Key=value?
- Users should be able to also add metadata to individual intervals in the log.
- Don't allow headers intermingled with interval histograms
- Don't allow duplicate headers

Spur of the moment strawman syntax:

```
!Version 2
!StartTime 123.456
#Comment
?Key=value
... normal interval histogram
```

Not sure what to do about metadata for an interval. Maybe re-use url query param encoding? But almost everyone gets that wrong, so perhaps it's best to steer clear of that mess.

marshallpierce:
Oh, and also: Fixes #59.

src/serialization/interval_log/mod.rs
```diff
@@ -0,0 +1,588 @@
...
+
+#[derive(PartialEq, Debug)]
+/// Represents one non-comment line in an interval log.
+///
+/// One thing to note is that the way your interval timestamps work can vary. If your log was
+/// written with a StartTime or BaseTime, that metadata will appear in header comments, and that
+/// will be represented by the iterator providing the corresponding variants here. The presence
+/// of those timestamps will affect how you should interpret the timestamps for individual
+/// intervals.
+#[allow(variant_size_differences)]
```
marshallpierce: Should we be worried about a 56-byte enum variant?

src/serialization/interval_log/mod.rs
```diff
@@ -0,0 +1,588 @@
...
+                text_buf: String::new(),
+                serialize_buf: Vec::new(),
+            },
+        }
+    }
+
+    /// Add a comment line.
+    ///
+    /// If you do silly things like write a comment with a newline character, you'll end up with
+    /// an un-parseable log file. Don't do that.
```
marshallpierce: Maybe panic? Or filter silently?

src/serialization/interval_log/mod.rs
```diff
@@ -0,0 +1,588 @@
...
+    pub fn tag(&self) -> Option<Tag<'a>> {
+        self.tag
+    }
+
+    /// Timestamp of the start of the interval in seconds.
+    ///
+    /// The timestamp may be absolute vs the epoch, or there may be a `StartTime` or `BaseTime` for
+    /// the log, in which case you may wish to consider this number as a delta vs those timestamps.
+    /// See the module-level documentation about timestamps.
+    pub fn start_timestamp(&self) -> f64 {
```
marshallpierce: Could add a method to return a `chrono` `NaiveDateTime`, but it didn't seem worth the dependency to do something that the consumer can just as easily do

src/serialization/interval_log/mod.rs
```diff
@@ -0,0 +1,588 @@
...
+    ///
+    /// `start_timestamp` is the time since the epoch in seconds. If you're using a StartTime or
+    /// BaseTime offset, you should instead use a delta since that time. See the discussion about
+    /// timestamps in the module-level documentation.
+    ///
+    /// `duration` is the duration of the interval in seconds.
+    ///
+    /// `tag` is an optional tag for this histogram.
+    ///
+    /// `max_value_divisor` is used to scale down the max value to something that may be more human
```
marshallpierce: Maybe this should be fixed for the entire interval log? It seems like it can't be a good idea to allow different divisors for different intervals.

src/serialization/interval_log/mod.rs
```diff
@@ -0,0 +1,587 @@
...
+//! use hdrsample::serialization::interval_log;
+//!
+//! let mut log = Vec::new();
+//! log.extend_from_slice(b"#I'm a comment\n");
+//! log.extend_from_slice(b"Tag=a,0.123,1.007,2.769,base64EncodedHisto\n");
+//! log.extend_from_slice(b"1.456,1.007,2.769,base64EncodedHisto\n");
+//! log.extend_from_slice(b"3.789,1.007,2.769,base64EncodedHisto\n");
+//! log.extend_from_slice(b"Tag=b,4.123,1.007,2.769,base64EncodedHisto\n");
+//! log.extend_from_slice(b"5.456,1.007,2.769,base64EncodedHisto\n");
+//! log.extend_from_slice(b"#Another comment\n");
```
jonhoo: This might be more readable with multi-line strings:
```rust
let mut log = "\
               #I'm a comment\n\
               Tag=a,0.123,1.007,2.769,base64EncodedHisto\n\
               1.456,1.007,2.769,base64EncodedHisto\n\
               3.789,1.007,2.769,base64EncodedHisto\n\
               Tag=b,4.123,1.007,2.769,base64EncodedHisto\n\
               5.456,1.007,2.769,base64EncodedHisto\n\
               #Another comment\n"
    .to_vec();
```

src/serialization/interval_log/mod.rs
```diff
@@ -0,0 +1,587 @@
...
+//!
+//! let iter = interval_log::IntervalLogIterator::new(&log);
+//!
+//! let count = iter.map(|r| r.unwrap())
+//!     // only look at intervals (which are the only non-comment lines in this log)
+//!     .filter_map(|e| match e {
+//!         interval_log::LogEntry::Interval(ilh) => Some(ilh),
```
jonhoo: I would match on `Ok(interval_log::LogEntry::Interval(ilh))` here to avoid the `unwrap` just above.

src/serialization/interval_log/mod.rs
```diff
@@ -0,0 +1,587 @@
...
+//! h.record(12345).unwrap();
+//!
+//! // limit scope of mutable borrow of `buf`
+//! {
+//!     let mut header_writer = interval_log::IntervalLogHeaderWriter::new(
+//!         &mut buf, &mut serializer);
+//!     header_writer.write_comment("Comments are great").unwrap();
+//!     header_writer.write_start_time(123.456789).unwrap();
+//!
+//!     let mut log_writer = header_writer.into_log_writer();
```
jonhoo: Haven't read further down yet, but is an `IntervalLogHeaderWriter` the only way to get an `IntervalLogWriter`? If so, I think the former should be given a different name.

src/serialization/interval_log/mod.rs
jonhoo: Without having read further down, I also don't see why the `IntervalLogIterator` yields a `Result`. What is the `Err` case? And wouldn't it be easier to use with the errors being another `LogEntry` type? Or does the iterator behave weirdly once an `Err` has been yielded?

src/serialization/interval_log/mod.rs
```diff
@@ -0,0 +1,587 @@
...
+
+/// Write headers for an interval log.
+///
+/// This type only allows writing comments and headers. Once you're done writing those things, use
+/// `into_log_writer()` to convert this into an `IntervalLogWriter`.
+pub struct IntervalLogHeaderWriter<'a, 'b, W: 'a + io::Write, S: 'b + Serializer> {
+    internal_writer: InternalLogWriter<'a, 'b, W, S>,
+}
+
+impl<'a, 'b, W: 'a + io::Write, S: 'b + Serializer> IntervalLogHeaderWriter<'a, 'b, W, S> {
```
jonhoo: I don't think the `IntervalLogHeaderWriter` should be parameterized on `S`. Instead, the call to `into_log_writer` should be. The header writes are all independent of the `Serializer` as far as I can tell?

src/serialization/interval_log/mod.rs
jonhoo: Or `.replace('\n', "\n#")`?

src/serialization/interval_log/mod.rs
```diff
@@ -0,0 +1,588 @@
...
+    ///
+    /// If you do silly things like write a comment with a newline character, you'll end up with
+    /// an un-parseable log file. Don't do that.
+    pub fn write_comment(&mut self, s: &str) -> io::Result<()> {
+        self.internal_writer.write_comment(s)
+    }
+
+    /// Write a StartTime log line. See the module-level documentation for more info.
+    ///
+    /// This should only be called once to avoid creating a confusing interval log.
```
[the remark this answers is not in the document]
jonhoo: Thinking about this some more, `IntervalLogHeaderWriter` really just seems like a builder to me. Maybe that's the way we should present it. That would allow us to only write things out when the user has committed to them, and then only write them once. Something like:
```rust
let mut ilb = IntervalLogBuilder::default().with_start_time(42);
ilb.add_comment("I love comments");
let mut logger = ilb.build(&mut writer, &mut serializer);
```
This has some other neat secondary effects such as headers being easy to re-use, fields that can only be set once only being written once, and more efficient writes (because they can be batch-written on `.build()`. We can bikeshed the exact name for `.build()`, but overall following the builder pattern as outlined in the [Rust API guidelines](https://aturon.github.io/ownership/builders.html) would be a good route to take.

src/serialization/interval_log/mod.rs
```diff
@@ -0,0 +1,587 @@
...
+    pub fn write_base_time(&mut self, seconds_since_epoch: f64) -> io::Result<()> {
+        self.internal_writer.write_fmt(format_args!(
+            "#[BaseTime: {:.3} (seconds since epoch)]\n",
+            seconds_since_epoch
+        ))
+    }
+
+    /// Once you're finished with headers, convert this into a log writer so you can write interval
+    /// histograms.
+    pub fn into_log_writer(self) -> IntervalLogWriter<'a, 'b, W, S> {
```
jonhoo: See notes above about making this more builder-like, and then only committing to a writer + serializer at this point (rather than when creating the header writer).

src/serialization/interval_log/mod.rs
```diff
@@ -0,0 +1,587 @@
...
+            internal_writer: self.internal_writer,
+        }
+    }
+}
+
+/// Writes interval histograms in an interval log.
+///
+/// This isn't created directly; start with an `IntervalLogHeaderWriter`. Once you've written the
+/// headers and ended up with an `IntervalLogWriter`, typical usage would be to write a histogram
+/// at regular intervals (e.g. once a second).
```
jonhoo: I think an example would be more useful here.

src/serialization/interval_log/mod.rs
```diff
@@ -0,0 +1,587 @@
...
+    }
+
+    /// Write an interval histogram.
+    ///
+    /// `start_timestamp` is the time since the epoch in seconds that measurements started being
+    /// recorded in this interval. If you're using a StartTime or BaseTime offset, you should
+    /// instead use a delta since that time. See the discussion about timestamps in the module-level
+    /// documentation.
+    ///
+    /// `duration` is the duration of the interval in seconds.
```
jonhoo: Any particular reason you didn't make this a `time::Duration`?

src/serialization/interval_log/mod.rs
jonhoo: Yeah, I think that sounds like a good idea. *And* it fits nicely with the proposed builder pattern!

src/serialization/interval_log/mod.rs
```diff
@@ -0,0 +1,587 @@
...
+    fn write_histogram<T: Counter>(
+        &mut self,
+        h: &Histogram<T>,
+        start_timestamp: f64,
+        duration: f64,
+        tag: Option<Tag>,
+        max_value_divisor: f64,
+    ) -> Result<(), IntervalLogWriterError<S::SerializeError>> {
+        self.serialize_buf.clear();
+        self.text_buf.clear();
```
jonhoo: Any particular reason you wanted `text_buf` and `serialize_buf` to be separate?

src/serialization/interval_log/mod.rs
```diff
@@ -0,0 +1,587 @@
...
+            self.text_buf,
+            start_timestamp,
+            duration,
+            h.max() as f64 / max_value_divisor // because the Java impl does it this way
+        )?;
+
+        self.text_buf.clear();
+        let _len = self.serializer
+            .serialize(h, &mut self.serialize_buf)
+            .map_err(|e| IntervalLogWriterError::SerializeError(e))?;
```
jonhoo: I would add
```rust
impl<E> From<E> for IntervalLogWriterError<E> {}
```
so that you can remove this `map_err`.

src/serialization/interval_log/mod.rs
```diff
@@ -0,0 +1,587 @@
...
+        }
+    }
+
+    /// Returns the tag contents as a str.
+    pub fn as_str(&self) -> &'a str {
+        self.0
+    }
+}
+
+impl<'a> ops::Deref for Tag<'a> {
```
jonhoo: You might want to `impl` `AsRef` instead here. Though not clear that it matters.

src/serialization/interval_log/mod.rs
jonhoo: Also, given that it might be entirely user-defined, I'm not sure we should.

src/serialization/interval_log/mod.rs
```diff
@@ -0,0 +1,587 @@
...
+        self.as_str()
+    }
+}
+
+/// An individual interval histogram parsed from an interval log.
+#[derive(PartialEq, Debug)]
+pub struct IntervalLogHistogram<'a> {
+    tag: Option<Tag<'a>>,
+    start_timestamp: f64,
+    duration: f64,
```
jonhoo: I still think this should be exposed as `time::Duration`.

src/serialization/interval_log/mod.rs
```diff
@@ -0,0 +1,587 @@
...
+    StartTime(f64),
+    /// Logs may include a BaseTime. If present, it represents seconds since the epoch.
+    BaseTime(f64),
+    /// An individual interval histogram.
+    Interval(IntervalLogHistogram<'a>),
+}
+
+/// Errors that occur when parsing an interval log.
+#[derive(Debug, PartialEq)]
+pub enum LogIteratorError {
```
jonhoo: This seems unfortunate, but makes total sense. There's some related discussion here: https://users.rust-lang.org/t/handling-errors-from-iterators/2551/20. I guess it's fine.

src/serialization/interval_log/mod.rs
```diff
@@ -0,0 +1,587 @@
...
+/// metadata before incurring the cost of deserialization. If you're not using any timestamp
+/// headers, or at least using them in straightforward ways, it is easy to accumulate the
+/// timestamp state you need. Since all the parsing is taken care of already, writing your own
+/// `HistogramLogReader` equivalent that fits the way your logs are assembled is just a couple of
+/// lines. (And if you're doing complex stuff, we probably wouldn't have built something that fits
+/// your quirky logs anyway!)
+///
+/// This parses from a slice representing the complete file because it made implementation easier
+/// (and also supports mmap'd files for maximum parsing speed). If parsing from a `Read` is
+/// important for your use case, open an issue about it.
```
jonhoo: This is an interesting point. I suspect parsing from a `Read` will be a very useful addition, and one that will be tricky to combine with the way the code is currently written, but I agree that we can deal with that on-demand.

src/serialization/interval_log/mod.rs
```diff
@@ -0,0 +1,587 @@
...
+    pub fn new(input: &'a [u8]) -> IntervalLogIterator<'a> {
+        IntervalLogIterator {
+            orig_len: input.len(),
+            input,
+        }
+    }
+}
+
+impl<'a> Iterator for IntervalLogIterator<'a> {
+    type Item = Result<LogEntry<'a>, LogIteratorError>;
```
jonhoo: I have some issues with this.
If I've understood the log format correctly, this iterator will *always* first yield at most one each of `StartTime` and `BaseTime`, and from that point forwards, only yield `LogEntry::Interval`. If that interpretation is correct, then it seems like this should really be an iterator that yields `IntervalLogHistogram<'a>`, with some kind of initial step that gives you the start/base time. That'd be much more convenient to work with, as it would obviate the need for that pesky `match`.

src/serialization/mod.rs
```diff
@@ -137,7 +144,7 @@
 //! // Make some histograms
 //! for _ in 0..num_histograms {
 //!     let mut h = Histogram::<u64>::new_with_bounds(1, u64::max_value(), 3).unwrap();
-//!     h.record_n(42, 7);
+//!     h.record_n(42, 7).unwrap();
```
jonhoo: I really want https://github.com/rust-lang/rfcs/pull/1937 so we can make this use `?`.

src/serialization/interval_log/mod.rs
marshallpierce: I'll make `write_histogram` use `time::Duration` too.

src/serialization/interval_log/mod.rs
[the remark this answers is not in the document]
marshallpierce: I like the builder approach.

src/serialization/interval_log/mod.rs
marshallpierce: Hmm... without a way to bound `E`, that would map every possible error type to `SerializeError`, which seems undesirable.

src/serialization/interval_log/mod.rs
marshallpierce: I don't think it will be too bad; we'll just have to read into a buffer, parse what we can, and grow the buffer if we get `Incomplete`. There's some streaming support in `nom`, but I didn't investigate too closely. More than I felt like tackling on top of learning `nom`. :)

src/serialization/interval_log/mod.rs
marshallpierce: OK, I'll add one once we get the builder API sorted out.

src/serialization/interval_log/mod.rs
jonhoo: I think you want [`char::is_control`](https://doc.rust-lang.org/std/primitive.char.html#method.is_control), which includes all the Unicode [control characters](http://www.fileformat.info/info/unicode/category/Cc/list.htm) (including newlines).

src/serialization/interval_log/mod.rs
jonhoo: I agree with all the above.

src/serialization/interval_log/mod.rs
jonhoo: But isn't that what the `IntervalLogWriterError` type already suggests?

src/serialization/interval_log/mod.rs
jonhoo: Yeah, this is really poorly documented. Reading a bit more though, I think implementing `Deref` might actually be the right thing after all, so leave it as-is.

src/serialization/interval_log/mod.rs
jonhoo: Hmm, I suppose... It seems unfortunate, because it makes the API a fair amount less ergonomic. I guess we could provide a refined iterator on top of this, whereas the reverse is not true, so leave it like this for now.

tests/interval_log.rs
```diff
@@ -250,10 +268,11 @@ mod tests {
             }
         }
 
+        println!("{}", ::std::str::from_utf8(&log_buf).unwrap());
```
jonhoo: I assume this is just leftover from debugging?

src/serialization/interval_log/mod.rs
marshallpierce: Well, there is the `IoError` variant, so it's not just a wrapper for serialization errors. Having everything else map to `SerializeError(E)` feels risky; wouldn't that make it possible to have a typo where some other `Result`-producing call gets a `?` appended to it, and presto now we're returning `SerializeError(SomeOtherUnrelatedError)`. 

`io::Error` seems like a reasonable thing for `From` because it's called many times and is unambiguous. I've never used an open-ended `From` like that so maybe my unease is unfounded, but it reminds me unpleasantly of Scala's implicits which made code hard to follow.

tests/interval_log.rs
marshallpierce: Doh!

src/serialization/interval_log/mod.rs
```diff
@@ -0,0 +1,596 @@
...
+
+#[derive(PartialEq, Debug)]
+/// Represents one non-comment line in an interval log.
+///
+/// One thing to note is that the way your interval timestamps work can vary. If your log was
+/// written with a StartTime or BaseTime, that metadata will appear in header comments, and that
+/// will be represented by the iterator providing the corresponding variants here. The presence
+/// of those timestamps will affect how you should interpret the timestamps for individual
+/// intervals. See the module-level documentation.
+#[allow(variant_size_differences)]
```
marshallpierce: Should we care about this? Not really sure I see a way around it

src/serialization/interval_log/mod.rs
jonhoo: What I mean is that `IntervalLogWriterError`'s one type argument is only used in the `SerializeError` variant. If we were to add some other wrapped error, that would be of a different type, which would change `IntervalLogWriterError` to take two type arguments `<E1, E2>`. In that case, an `impl` of `From` wouldn't make much sense without further bounds. But in the current case, if you have an error and want to produce some `IntervalLogWriterError<E>`, then that works exactly when that error is of type `E`. Which is what the above `From` would work for.

src/serialization/interval_log/mod.rs
jonhoo: I thought `variant_size_differences` was allowed by default? I think it's fine for this `enum` to have a somewhat large size variant, though it is also another reason to make the iterator only yield `IntervalLogHistogram` (as the enum would no longer be necessary).

src/serialization/interval_log/mod.rs
marshallpierce: I tightened up the bounds a bit:
```
pub enum IntervalLogWriterError<S: Serializer> {
    /// Histogram serialization failed.
    SerializeError(S::SerializeError),
    /// An i/o error occurred.
    IoError(io::ErrorKind),
}
```

Nonetheless, the compiler gets grumpy:
> error[E0119]: conflicting implementations of trait `std::convert::From<serialization::interval_log::IntervalLogWriterError<_>>` for type `serialization::interval_log::IntervalLogWriterError<_>

jonhoo: What code it it's complaining on?

marshallpierce: I've extracted it into this playground: https://play.rust-lang.org/?gist=d6826495d99f9718c0980197a0fe14d3&version=stable

I think what the compiler is getting at is that you could have a `Serializer` whose `SerializeError` was `io::Error`, and then there would be no way to know which `impl` of `From` to use.

src/serialization/interval_log/mod.rs
marshallpierce: They are, but they're turned off in the crate-wide `deny` statement. The variant is only large because `IntervalLogHistogram` is large, though, so removing the enum isn't really going to help the "copying fat structs around" issue (though I think it's unlikely to be a performance issue).

src/serialization/interval_log/mod.rs
jonhoo: Ah, yes, that makes sense. Oh well, leave it as-is I guess.

src/serialization/interval_log/mod.rs
```diff
@@ -0,0 +1,667 @@
...
+    ///
+    /// This can be called multiple times, but only the value for the most recent invocation will
+    /// be written.
+    pub fn with_max_value_divisor(&mut self, max_value_divisor: f64) -> &mut Self {
+        self.max_value_divisor = max_value_divisor;
+        self
+    }
+
+    /// Build a LogWriter and apply any configured headers.
+    pub fn build_with<'a, 'b, W: 'a + io::Write, S: 'b + Serializer>(
```
marshallpierce: This is pretty much the opposite of what https://aturon.github.io/ownership/builders.html suggests: we're taking the required things at the end, not the beginning, but being able to use the same setup on multiple writers or serializers seems hypothetically useful, and this API does not offend me otherwise.

Let the name bikeshedding begin! :P

src/serialization/interval_log/mod.rs
```diff
@@ -0,0 +1,667 @@
...
+        if let Some(bt) = self.base_time {
+            internal_writer.write_fmt(format_args!(
+                "#[BaseTime: {:.3} (seconds since epoch)]\n",
+                bt
+            ))?;
+        }
+
+        // The Java impl doesn't write a comment for this but it's confusing to silently modify the
+        // max value without leaving a trace
+        if self.max_value_divisor != 1.0_f64 {
```
marshallpierce: Straying off the reservation here but it is so easy to misinterpret a log (as a human) without this

src/serialization/interval_log/mod.rs
```diff
@@ -170,69 +169,138 @@ use nom::{double, line_ending, not_line_ending, IResult};
...
     ///
-    /// This should only be called once to avoid creating a confusing interval log.
-    pub fn write_start_time(&mut self, seconds_since_epoch: f64) -> io::Result<()> {
-        self.internal_writer.write_fmt(format_args!(
-            "#[StartTime: {:.3} (seconds since epoch)]\n",
-            seconds_since_epoch
-        ))
+    /// This can be called multiple times, but only the value for the most recent invocation will
+    /// be written.
+    pub fn with_start_time(&mut self, seconds_since_epoch: f64) -> &mut Self {
```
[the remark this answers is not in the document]
marshallpierce: Hmm... perhaps. That would mean we'd have to do something about handling the error case for `duration_since()`, though. We could just take the `Duration` instead of the `SystemTime` to push that decision off to the caller, but still have something more "time-flavored" than a lowly `f64`.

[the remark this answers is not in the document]
jonhoo:
```rust
let since_epoch = match time.duration_since(UNIX_EPOCH) {
    Ok(d) => dur_to_f64(d),
    Err(ste) => -1 * dur_to_f64(ste.duration())
};
```
?

src/serialization/interval_log/mod.rs
jonhoo: Eh, I'm okay with this approach.

I actually don't mind `build_with`. I think I'd be partial to a name that fits better in context though. How about `begin_log_with`?

src/serialization/interval_log/mod.rs
jonhoo: I like it.

src/serialization/interval_log/mod.rs
marshallpierce: Sure, WFM.

src/serialization/interval_log/mod.rs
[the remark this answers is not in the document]
marshallpierce: Yeah, that should work, and we're very unlikely to hit the Err case, though I wonder if maybe we should panic because your timekeeping is clearly hosed if your system thinks it's before 1970?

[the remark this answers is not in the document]
marshallpierce: Doh -- if we accept `SystemTime`, we can't have repeatable tests because you can't construct an arbitrary `SystemTime`.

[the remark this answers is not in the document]
jonhoo: Actually, you can, by adding to `UNIX_EPOCH` ;)
As for panicking, I'd rather avoid it given that we don't have to.

src/serialization/interval_log/mod.rs
```diff
@@ -0,0 +1,667 @@
...
+    ) -> Result<IntervalLogWriter<'a, 'b, W, S>, io::Error> {
+        let mut internal_writer = InternalLogWriter {
+            writer,
+            serializer,
+            text_buf: String::new(),
+            serialize_buf: Vec::new(),
+            max_value_divisor: self.max_value_divisor,
+        };
+
+        for c in self.comments.iter() {
```
jonhoo: Just FYI, you can use `for c in &self.comments` here.

src/serialization/interval_log/mod.rs
```diff
@@ -0,0 +1,667 @@
...
+            start_time: None,
+            base_time: None,
+            max_value_divisor: 1.0,
+        }
+    }
+
+    /// Add a comment line to be written when the writer is built.
+    ///
+    /// If you do silly things like write a comment with a newline character, you'll end up with
+    /// an un-parseable log file. Don't do that.
```
jonhoo: Just want to make sure that https://github.com/jonhoo/hdrsample/pull/70#discussion_r148692715 doesn't get lost in these changes.

marshallpierce: Yep, got it tracked in the opening comment on the PR. This is one of the things I prefer about BitBucket: PR tasks!

src/serialization/interval_log/mod.rs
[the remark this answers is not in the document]
marshallpierce: Ah, good point. In that case I'm 👍 because I can get my testing done *and* we can always provide another way to set the times if `SystemTime` turns out to be awkward for users.

src/serialization/interval_log/mod.rs
```diff
@@ -0,0 +1,667 @@
...
+    /// instead use a delta since that time. See the discussion about timestamps in the module-level
+    /// documentation.
+    ///
+    /// `duration` is the duration of the interval in seconds.
+    ///
+    /// `tag` is an optional tag for this histogram.
+    pub fn write_histogram<T: Counter>(
+        &mut self,
+        h: &Histogram<T>,
+        start_timestamp: f64,
```
marshallpierce: I wonder if this should maybe be a `Duration` as well? 🤔 Conceivably, that's how it would be used: a `Duration` since either `UNIX_EPOCH` or whatever the chosen `StartTime`/`BaseTime` was.

jonhoo: Oh, yeah, that's a good idea!

src/serialization/interval_log/mod.rs
```diff
@@ -0,0 +1,683 @@
...
+///
+/// One thing to note is that the way your interval timestamps work can vary. If your log was
+/// written with a StartTime or BaseTime, that metadata will appear in header comments, and that
+/// will be represented by the iterator providing the corresponding variants here. The presence
+/// of those timestamps will affect how you should interpret the timestamps for individual
+/// intervals. See the module-level documentation.
+#[allow(variant_size_differences)]
+pub enum LogEntry<'a> {
+    /// Logs may include a StartTime. If present, it represents seconds since the epoch.
+    StartTime(f64),
```
marshallpierce: And I suppose this should be `Duration` then too. More expensive to parse into `Duration` but these should be rare relative to histograms, so it's probably not an issue.

src/serialization/interval_log/mod.rs
marshallpierce: I tightened up parsing to only accept newlines as a line ending (because `\r` should be shunned), and parsing, which was already quite fast, sped up over 3x!

Anyway, it seems to work fine if we only transform `\n` into newline-plus-# and leave everything else as-is. Is there a particular reason to turn any control character into a subsequent line of comments?

src/serialization/interval_log/mod.rs
marshallpierce: I take it back, that should be `SystemTime` I suppose.

marshallpierce: Hm, now I'm not so sure. There's not really much useful you can do with a `SystemTime` except turn it back into a `Duration` based on `UNIX_EPOCH`, which would force the user to do another round of error checking on the `Result`. So, it seems probably more desirable to give them the `Duration` directly. The path from `Duration` to `chrono::naive::NaiveDateTime` and friends is then pretty smooth.

src/serialization/interval_log/mod.rs
```diff
@@ -365,7 +366,11 @@ impl<'a, 'b, W: 'a + io::Write, S: 'b + Serializer> InternalLogWriter<'a, 'b, W,
     }
 
     fn write_comment(&mut self, s: &str) -> io::Result<()> {
-        write!(self.writer, "#{}\n", s)
+        for l in s.split('\n') {
+            write!(self.writer, "#{}\n", l)?;
```
jonhoo: Do you not still have to escape control characters? `\r` for example?

marshallpierce: Not unless we start treating `\r` as a line separator. I added a test that has every control character in a comment. It gets a new line with a leading `#` for the `\n` but all others are left alone, and it parses fine.

src/serialization/interval_log/mod.rs
```diff
@@ -681,5 +674,44 @@ named!(legend<&[u8], ()>,
 
 named!(ignored_line<&[u8], ()>, alt!(comment_line | legend));
 
+fn fract_sec_duration(input: &[u8]) -> IResult<&[u8], time::Duration> {
+    match fract_sec_tuple(input) {
+        IResult::Done(rest, data) => {
+            let (secs, nanos_str) = data;
+
+            // only read up to 9 digits since we can only support nanos, not smaller precision
```
jonhoo: "only" heh

jonhoo:
cfec24 fixes the last task "Better type for reading StartTime/BaseTime", right? I think I'm happy to merge.

marshallpierce:
Yep I think this is good to go!

jonhoo:
Nice! Good work.

PR #71, opened 2017-12-08
Clarify interval log semantics.

marshallpierce:
See https://gitter.im/HdrHistogram/HdrHistogram?at=5a1a731771ad3f87362ebd42.

Who knows if anyone will ever read all of this, but the lack of clarity was bothering me. ;)

### DIFF
--- a/src/serialization/interval_log/mod.rs
+++ b/src/serialization/interval_log/mod.rs
@@ -12,58 +12,102 @@
 //! The intervals in the log should be ordered by start timestamp. It's possible to write (and
 //! parse) logs with intervals in any order, but the expectation is that they will be sorted.
 //!
-//! To parse a log, see `IntervalLogIterator`. To write a log, see `IntervalLogHeaderWriter`.
+//! To parse a log, see `IntervalLogIterator`. To write a log, see `IntervalLogWriterBuilder`.
 //!
 //! # Timestamps
 //!
 //! Each interval has a timestamp in seconds associated with it. However, it's not necessarily as
-//! simple as just interpreting the number as seconds since the epoch.
+//! simple as just interpreting the number as seconds since the epoch. There are two optional pieces
+//! of header metadata: "StartTime" and "BaseTime". Neither, one, or both of these may be present.
+//! It is possible to have multiple StartTime or BaseTime entries in the log, or even interleaved
+//! with interval histograms, but that is confusing, so this API prevents you from doing so.
 //!
-//! There are two optional pieces of header metadata: "StartTime" and "BaseTime". Neither, one, or
-//! both of these may be present. It is possible to have multiple StartTime or BaseTime entries in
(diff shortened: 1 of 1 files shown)


src/serialization/interval_log/mod.rs
```diff
@@ -12,58 +12,102 @@
...
+//! some cases log files are consumed by humans too.
+//!
+//! While BaseTime is used to redefine per-interval timestamps for human readability, StartTime
+//! provides information about when the process that is generating the log reached some "start"
+//! condition. It's frequently the case that intervals will start some seconds after the process
+//! started, whether due to initial warmup before a benchmark or because it just takes a
+//! while to start up. If this is relevant to your workload, use StartTime to record the actual
+//! process start time (or other relevant "start" condition, like when a benchmark begins on an
+//! already long-running process). You could then use this when processing a log to more accurately
+//! plot interval data over time.
```
jonhoo: Are the per-interval timestamps then relative to `StartTime`?

jonhoo: From reading the example below, I take it they aren't (that is, they are *always* relative to `BaseTime` or UNIX epoch). We might want to clarify that here.

src/serialization/interval_log/mod.rs
```diff
@@ -12,58 +12,102 @@
...
+//!
+//! - Neither StartTime nor BaseTime are present: interval timestamps are interpreted as seconds
+//! since the epoch. The first interval's timestamp is stored to the StartTime field.
 //! - StartTime is present: StartTime is a number of seconds since epoch, and interval timestamps
-//! should be interpreted as deltas that could be added to StartTime if seconds since epoch for each
-//! interval is needed.
-//! - BaseTime is present: same as the case where StartTime is present. It's seconds since epoch,
-//! with interval timestamps as deltas.
+//! may be interpreted as deltas to be added to StartTime or as "absolute" Unix time depending on a
+//! heuristic. In other words, the heuristic chooses between setting the effective BaseTime to 0 or
```
jonhoo: Is this heuristic straightforward to explain?

src/serialization/interval_log/mod.rs
```diff
@@ -12,58 +12,102 @@
...
+//! to StartTime.
+//! - BaseTime is present: BaseTime is a number of seconds since epoch, and interval timestamps are
+//! interpreted as deltas. The first interval's timestamp is stored to the StartTime field.
```
jonhoo: Hmm.. So `StartTime` is also relative to `BaseTime`? Sort of?

marshallpierce: This kind of thing is why I didn't want to make a higher level iterator that's an equivalent to HistogramLogReader until we get some clarity from users about what parts of its behavior they actually use. :)

jonhoo:
LGTM!

PR #74, opened 2017-12-21
Add saturating_record for clamped recording

### LINKED ISSUE #72: Add method for clamped recording
In some benchmarking scenarios, I am running into extreme outliers, often due to a bug in the system or due to hitting a scaling wall. When these happen, my benchmark binary crashes, since I record using something like:
```rust
let us = elapsed.as_secs() * 1_000_000 + elapsed.subsec_nanos() as u64 / 1_000;
hist.record(us).unwrap();
```
Operations generally take ~10µs, but the outliers can easily grow to tens of *seconds*. So even with liberal bounds, the record will fail. When I hit these outliers, I don't actually care about *how* much of an outlier they are, since I don't expect any "regular" samples beyond, say, ~1ms. In this case, what I really want is the ability to record values clamped to the range of the histogram (which should then never fail). I end up writing code like:
```rust
if hist.record(us).is_err() {
   let m = hist.high();
    hist.record(m).unwrap();
}
```
It'd be really nice if I could instead write:
```rust
hist.record_clamped(us);
```

@marshallpierce thoughts?

jonhoo:
Fixes #72.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -817,7 +817,8 @@ impl<T: Counter> Histogram<T> {
     ///
     /// This method cannot fail, as any values that are too small or too large to be tracked will
     /// automatically be clamed to be in range. Be aware that this *will* hide extreme outliers
-    /// from the resulting histogram without warning.
+    /// from the resulting histogram without warning. Since the values are clamped, the histogram
+    /// will also not be resized to accomodate the value, even if auto-resize is enabled.
     pub fn saturating_record(&mut self, value: u64) {
         self.saturating_record_n(value, T::one())
     }
@@ -838,7 +839,8 @@ impl<T: Counter> Histogram<T> {
     ///
     /// This method cannot fail, as values that are too small or too large to be recorded will
     /// automatically be clamed to be in range. Be aware that this *will* hide extreme outliers
-    /// from the resulting histogram without warning.
+    /// from the resulting histogram without warning. Since the values are clamped, the histogram
+    /// will also not be resized to accomodate the value, even if auto-resize is enabled.
     pub fn saturating_record_n(&mut self, value: u64, count: T) {
         self.record_n_inner(value, count, true).unwrap()
     }
@@ -861,8 +863,8 @@ impl<T: Counter> Histogram<T> {
                     self.lowest_discernible_value
                 };
 
-                // unwrap must succeed since low and high are always representable
-                let c = self.mut_at(value).unwrap();
+                let c = self.mut_at(value)
+                    .expect("unwrap must succeed since low and high are always representable");
                 *c = c.saturating_add(count);
             } else if !self.auto_resize {
                 return Err(RecordError::ValueOutOfRangeResizeDisabled);
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -813,12 +813,37 @@ impl<T: Counter> Histogram<T> {
         self.record_n(value, T::one())
     }
 
+    /// Record `value` in the histogram, clamped to the range of the histogram.
+    ///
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -813,12 +813,37 @@ impl<T: Counter> Histogram<T> {
         self.record_n(value, T::one())
     }
 
+    /// Record `value` in the histogram, clamped to the range of the histogram.
+    ///
+    /// This method cannot fail, as any values that are too small or too large to be tracked will
+    /// automatically be clamed to be in range. Be aware that this *will* hide extreme outliers
+    /// from the resulting histogram without warning.
```
marshallpierce: We should also notify the user that this will *not* resize the histogram even if it is an auto-resize histogram -- we are clamping to the current bounds, not what they might be expanded to.

src/lib.rs
```diff
@@ -827,21 +852,33 @@ impl<T: Counter> Histogram<T> {
...
+                value = if value > self.highest_trackable_value {
+                    self.highest_trackable_value
+                } else {
+                    // must be smaller than the lowest_discernible_value, since self.mut_at(value)
+                    // failed, and it's not too large (per above).
+                    self.lowest_discernible_value
+                };
+
+                // unwrap must succeed since low and high are always representable
+                let c = self.mut_at(value).unwrap();
```
marshallpierce: I'm not sure about the convention on these things but is it perhaps preferable to have this be an `expect()` with the info in the comments as the argument?

tests/histogram.rs
```diff
@@ -83,6 +83,15 @@ fn record_past_trackable_max() {
     assert!(h.record(3 * TRACKABLE_MAX).is_err());
 }
 
+#[test]
+fn saturating_record_past_trackable_max() {
+    let mut h = Histogram::<u64>::new_with_max(TRACKABLE_MAX, SIGFIG).unwrap();
+    h.saturating_record(3 * TRACKABLE_MAX);
```
marshallpierce: I'd feel better if this also had asserts for clamping on the low end, and a value in the middle that wasn't clamped.

jonhoo: Good point!

However, I'm running into issues. The following test fails:
```rust
let mut h = Histogram::<u64>::new_with_bounds(1000, TRACKABLE_MAX, SIGFIG).unwrap();
h.saturating_record(1); // clamped below
h.saturating_record(1000*1000); // not clamped
h.saturating_record(3 * TRACKABLE_MAX); // clamped above
assert_eq!(h.count_at(h.low()), 1);
```
`h.low()` correctly returns 1000, but `iter_recorded` gives a first sample of
```rust
IterationValue { value_iterated_to: 511, quantile: 0.3333333333333333, quantile_iterated_to: 0.3333333333333333, count_at_value: 1, count_since_last_iteration: 1 }
```
This suggests that there are bins below `h.low()`, which isn't entirely surprising if you know the internal representation of HdrHistogram, but will likely be unexpected for many users. I wonder how we might go about fixing that...

jonhoo: The way to fix this for now is to create the histogram with a min bound of 512, and then check that a recorded value of `1` actually adds to the count at 511, but I'd prefer being able to query the 511 from the histogram. I think this suggests to me that `low()` currently returns the wrong value. It should return the *actual* low, not the one passed in by the user.

marshallpierce: I think you're right -- you configure the histogram with what you NEED to be able to represent, and it will do that, but may also support other values. That semantic makes sense to me, so we might as well expose the actual limits.

I wonder if we should record the actual limits rather than the requested range when serializing? That's arguably more useful when determining compatibility for, say, deserializing a histogram *into* another histogram...

PR #76, opened 2017-12-30
update rug to 0.9, and use rug more idiomatically

tspiteri:
This commit:

* updates the rug dependency to the latest version
* only uses the required features of rug to save rug compilation time as it doesn't have to compile the float and complex libraries
* uses rug more idiomatically

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,7 +36,7 @@ base64 = { version = "0.7.0", optional = true }
 
 [dev-dependencies]
 rand = "0.3.15"
-rug = "0.6.0"
+rug = "0.9.1"
 ieee754 = "0.2.2"
 clap = "2.26.2"
 
--- a/tests/quantile.rs
+++ b/tests/quantile.rs
@@ -9,7 +9,7 @@ use rand::Rng;
 use rand::distributions::range::Range;
 use rand::distributions::IndependentSample;
 use ieee754::Ieee754;
-use rug::{Integer, Rational};
+use rug::Rational;
 
 #[test]
 fn value_at_quantile_internal_count_exceeds_bucket_type() {
@@ -215,9 +215,7 @@ fn value_at_quantile_matches_quantile_at_each_value_sequence_values() {
         assert_eq!(length, h.len());
 
         for v in 1..(length + 1) {
-            let quantile = (Rational::from(Integer::from(v as u64))
-                / Rational::from(Integer::from(length as u64)))
-                .to_f64();
+            let quantile = Rational::from((v as u64, length as u64)).to_f64();
             let calculated_value = h.value_at_quantile(quantile);
             if !h.equivalent(v, calculated_value) {
                 println!(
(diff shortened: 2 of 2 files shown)


jonhoo:
I have half a mind to leave the `rug` dependency on with default features, just because then users of the crate who also use `rug` won't have to compile it twice. But not sure what the right policy is there. Otherwise it looks good barring the inline question for @marshallpierce

tests/quantile.rs
```diff
@@ -373,16 +369,8 @@ impl<'a, R: Rng + 'a> Iterator for RandomMaxIter<'a, R> {
...
-
-    // emulate ceil()
-    if product == Rational::from(prod_as_int.clone()) {
-        return prod_as_int.to_u64().unwrap();
-    } else {
-        // to_integer()'s rounding down has chopped off a fractional part
-        return prod_as_int.to_u64().unwrap() + 1;
-    }
+    let product = Rational::from_f64(quantile).unwrap() * Rational::from(count);
+    product.ceil().to_u64().unwrap()
```
jonhoo: @marshallpierce I know you've had a fair bit of time to play with these corner cases. Do you think rug's `ceil` does what you expect?

tspiteri:
The `rug` crate is only a dev-dependency in this case, so users of the `hdrsample` crate do not have to compile `rug` at all unless they are using it directly. Besides, if

* crate `r` has features `rat` and `flo`,
* crate `h` depends on `r` using only `rat`, and
* crate `u` depends on `h`, and depends on `r` directly using only `flo`, then

crate `r` will be compiled once with both features `rat` and `flo` enabled.

tests/quantile.rs
marshallpierce: It looks fine; I can't recall why I avoided using it to begin with.

Cargo.toml
```diff
@@ -36,7 +36,7 @@ base64 = { version = "0.7.0", optional = true }
 
 [dev-dependencies]
 rand = "0.3.15"
-rug = "0.6.0"
+rug = { version = "0.9.0", default-features = false, features = ["rational"] }
```
marshallpierce: Looks like 0.9.1 is the new hotness.

marshallpierce:
I'm also unclear on why we should bother with changing the features enabled for `rug` -- only people who are running `hdrsample`'s tests would need to compile it, right?

tspiteri:
Yes, it's people who are running the tests than need to compile it, I thought saving twenty seconds there couldn't hurt.

Cargo.toml
tspiteri: I have a habit of using the lowest version that is semver-compatible with the latest version and that provides all the required functionality.(0.9.0 is semver-compatible with 0.9.1, so cargo will get 0.9.1 anyway, and it has all the used functionality.)

marshallpierce: If cargo will resolve it to 0.9.1, let's specify 0.9.1. I view the default cargo behavior, when combined with the default advice of not checking in Cargo.lock, as a terrible mis-feature. I forget all the time that I have to add a `^` to make Cargo do what I want or I would have had it `^`'d from the beginning. :(

marshallpierce:
OK. I don't really care one way or another; one-time compile costs don't really bother me (as opposed to compilation costs that I have to pay every time I run the tests) but I don't mind removing them if it bothers other folks.

Cargo.toml
tspiteri: I'll change it to 0.9.1 then, and I'll also remove the features stuff. As you said below, one time costs during testing are not a real issue, so the feature stuff is basically noise in the Cargo.toml file.

jonhoo:
Thanks @tspiteri!

PR #82, opened 2019-06-19
Add initial support for multi-threaded recording.

### LINKED ISSUE #51: Add Recorder support
[Recorder](https://hdrhistogram.github.io/HdrHistogram/JavaDoc/org/HdrHistogram/Recorder.html) is really useful in real-world situations: one (or more) threads recording, with another thread occasionally inspecting the resulting histograms.

[synchronoise](https://github.com/QuietMisdreavus/synchronoise) has an implementation of [WriterReaderPhaser](https://stuff-gil-says.blogspot.com/2014/11/writerreaderphaser-story-about-new.html) that may prove useful.

jonhoo:
This PR introduces a starting point for multi-threaded histogram recorders ala those supported by [Java's `Recorder`](https://static.javadoc.io/org.hdrhistogram/HdrHistogram/2.1.11/org/HdrHistogram/Recorder.html). The `Recorder` implementation in this PR is somewhat different from the Java version though, as outlined below.

The basic design here is as follows: each `Recorder` keeps a regular (i.e., non-atomic) local `Histogram`. All recorded samples go to that histogram, so common-case recording should be very fast. On every write, the `Recorder` also checks a shared atomic numeric counter, `phase`, which indicates which phase the (one) reader is currently in. Normally, this phase is static, so the read will incur not cross-core traffic, and is wait-free. The reader (`SyncHistogram`) also keeps a regular (non-atomic) local `Histogram`, and all reads go to that histogram, incurring no synchronization cost.

As discussed thus far, the reader never sees any recorded samples, which seems problematic. The reader therefore has a `refresh` operation available to them. When `refresh` is called, the next write to each `Recorder` triggers a "phase shift", where all samples are merged into the reader's map. Each `Recorder` swaps out their histogram with an empty one, and sends the full one over an unbounded channel that the reader reads from. Once the reader has received histograms from each `Recorder` instance, `refresh` returns, and the reader observes all the recorded samples.

The assumption here is that calls to `refresh` are rare compared to calls to the `record` methods. `Recorder`s _only_ pay a cost when a reader wants up-to-date samples.

The current design requires _all_ `Recorder` instances to perform writes continuously to avoid blocking a `refresh`. The `Recorder::idle` method provides one escape-hatch for this, but it's still pretty unwieldy. `SyncHistogram::refresh_timeout` gives readers a way to avoid blocking perpetually, but is a suboptimal solution.

Unfortunately, it is not clear to me at the moment how we avoid this. Sadly I think it would be quite expensive to allow the reader to forcibly merge a `Recorder`'s state, even if the reader could somehow detect that a given `Recorder` is idle. It basically requires either double-buffering (each `Recorder` keeps two histograms, and the reader forces them to swap between them), or a histogram with concurrent (atomic) counter values, neither of which are great. It might be that you can get to a good place by having each `Recorder` keep a `ConcurrentHistogram` (i.e., one that uses atomics for every counter), but that is also _significantly_ more work, so I'll leave that as future work.

The current implementation also does not interact with intervals in any meaningful way, which I _believe_ the Java implementation does.

Given all this, I'm inclined to land this first iteration as a starting point. The API seems reasonable, and we could always improve the implementation later. The `Recorder::idle`, and `SyncHistogram::phase_timeout` should also allow users to not have to worry too much about blocking issues. It basically gives bounded staleness, and lets the user control the bounds. I _think_ the external API is also what we'll want for `Recorder` even with a more elaborate underlying implementation, so I believe it's fine to use the name `Recorder`. Alternative names for methods and for `SyncHistogram` are welcome.

Fixes #51.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,7 @@ maintenance = { status = "passively-maintained" }
 [features]
 bench_private = [] # for enabling nightly-only feature(test) on the main crate to allow benchmarking private code
 serialization = [ "flate2", "nom", "base64" ]
-sync = []
+sync = [ "crossbeam-channel" ]
 default = [ "serialization", "sync" ]
 
 [dependencies]
@@ -36,6 +36,7 @@ byteorder = "1.0.0"
 flate2 = { version = "1.0", optional = true }
 nom = { version = "^4.0", optional = true }
 base64 = { version = "0.10", optional = true }
+crossbeam-channel = { version = "0.3", optional = true }
 
 [dev-dependencies]
 rand = "0.6"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,8 @@ maintenance = { status = "passively-maintained" }
 [features]
 bench_private = [] # for enabling nightly-only feature(test) on the main crate to allow benchmarking private code
 serialization = [ "flate2", "nom", "base64" ]
-default = [ "serialization" ]
+sync = []
+default = [ "serialization", "sync" ]
 
 [dependencies]
 num-traits = "0.2"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1845,4 +1845,5 @@ pub mod errors;
 pub mod serialization;
 pub use self::core::counter::*;
 pub use errors::*;
+#[cfg(feature = "sync")]
 pub mod sync;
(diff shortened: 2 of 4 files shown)


marshallpierce:
Oof, I really wish I had the time to dig into this; it looks quite interesting. A from-arms-length :+1:.

jonhoo:
Coverage currently fails due to https://github.com/xd009642/tarpaulin/issues/190

jonhoo:
Actually, I'd be curious to hear your thoughts on this @tobz

PR #86, opened 2019-11-30
implement fmt::Display and std::error::Error for all internal errors

### LINKED ISSUE #81: `fmt::Display` trait for errors
Hi and thanks for a great library!

As the topic states, could it be possible to implement `Display` or even `Error` trait(s) for the crate's error types? I believe it will improve user experience and generally make it easier to incorporate those errors into the end-user's error type system :)

Thanks!

AnderEnder:
Closes #81

Initial implementations:
- error messages were taken from the first sentences of documentation strings.
- underlined **io::ErrorKind** does not have **Display** and **to_string**, while **as_str** is available, but private, so **Debug** representation was taken.

Proposition:
- replace ErrorKind with original **io::Error**, which already has **Display** implementation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,15 @@
+# Changelog
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
+
+## [Unreleased]
+### Added
+- `std::error:Error` implementation for internal Error types: `CreationError`, `AdditionError`, `SubtractionError`, `RecordError`, `UsizeTypeTooSmall`, `DeserializeError`, `IntervalLogWriterError`, `V2DeflateSerializeError`, `V2SerializeError`
+- Changelog
+
+### Changed
+- `DeserializeError` and `V2DeflateSerializeError` lost derived traits: `PartialEq`, `Eq`, `Clone`, `Copy`
+- Inner error type from `std::io::ErrorKind` to `std::io::Error` in types: `DeserializeError`, `V2DeflateSerializeError`, `V2SerializeError`, `IntervalLogWriterError` to support `Display`.
+
+### Removed
--- a/src/serialization/deserializer.rs
+++ b/src/serialization/deserializer.rs
@@ -3,15 +3,15 @@ use crate::{Counter, Histogram, RestatState};
 use byteorder::{BigEndian, ReadBytesExt};
 use flate2::read::ZlibDecoder;
 use num_traits::ToPrimitive;
-use std::io::{self, Cursor, ErrorKind, Read};
+use std::io::{self, Cursor, Read};
 use std::marker::PhantomData;
 use std::{self, error, fmt};
 
 /// Errors that can happen during deserialization.
-#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+#[derive(Debug)]
 pub enum DeserializeError {
     /// An i/o operation failed.
-    IoError(ErrorKind),
+    IoError(io::Error),
     /// The cookie (first 4 bytes) did not match that for any supported format.
     InvalidCookie,
     /// The histogram uses features that this implementation doesn't support (yet), so it cannot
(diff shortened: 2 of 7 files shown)


jonhoo:
Love it!

AnderEnder:
@jonhoo,
What do you think about my proposition to replace inner **ErrorKind** with **io::Error**?

jonhoo:
Yup, that seems reasonable to me.

AnderEnder:
A side effect of switching from **io::ErrorKind** to **io::Error** is loosing derive traits: **PartialEq**, **Eq**, **Clone**, **Copy**.
 
Only one test was affected, so I replaced  the comparison from Errors to Strings.

jonhoo:
Ah, that's a little awkward, but I guess this is a backwards-incompatible change anyway.. I think we should start keeping a changelog for changes like this. Could you add a file like [this one](https://raw.githubusercontent.com/jonhoo/inferno/master/CHANGELOG.md) and add this change under "Changed" in `[Unreleased]` with a link to this PR?

src/serialization/deserializer.rs
```diff
@@ -30,10 +30,44 @@ pub enum DeserializeError {
...
+            ),
+            DeserializeError::EncodedArrayTooLong => write!(
+                f,
+                "The encoded array is longer than it should be for the histogram's value range"
+            ),
+        }
+    }
+}
+
+impl error::Error for DeserializeError {}
```
jonhoo: This should override the implementation of [`Error::source`](https://doc.rust-lang.org/std/error/trait.Error.html#method.source).

src/serialization/v2_deflate_serializer.rs
```diff
@@ -5,24 +5,39 @@ use crate::Histogram;
...
+                write!(f, "The underlying serialization failed: {}", e)
+            }
+            V2DeflateSerializeError::IoError(e) => {
+                write!(f, "The underlying serialization failed: {}", e)
+            }
+        }
+    }
+}
+
+impl error::Error for V2DeflateSerializeError {}
```
jonhoo: This should override the implementation of [`Error::source`](https://doc.rust-lang.org/std/error/trait.Error.html#method.source).

src/serialization/v2_serializer.rs
```diff
@@ -14,15 +14,32 @@ pub enum V2SerializeError {
...
+            ),
+            V2SerializeError::UsizeTypeTooSmall => {
+                write!(f, "Internal calculations cannot be represented in `usize`")
+            }
+            V2SerializeError::IoError(e) => write!(f, "An i/o operation failed: {}", e),
+        }
+    }
+}
+
+impl error::Error for V2SerializeError {}
```
jonhoo: This should override the implementation of [`Error::source`](https://doc.rust-lang.org/std/error/trait.Error.html#method.source).

src/serialization/interval_log/mod.rs
```diff
@@ -407,12 +407,12 @@ pub enum IntervalLogWriterError<E> {
     /// Histogram serialization failed.
     SerializeError(E),
     /// An i/o error occurred.
-    IoError(io::ErrorKind),
+    IoError(io::Error),
```
jonhoo: Ooops, looks like I missed one: this should also override `source`.

AnderEnder: done

CHANGELOG.md
```diff
@@ -0,0 +1,15 @@
...
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
+
+## [Unreleased]
+### Added
+- `std::error:Error` implementation for internal Error types: `CreationError`, `AdditionError`, `SubtractionError`, `RecordError`, `UsizeTypeTooSmall`, `DeserializeError`, `IntervalLogWriterError`, `V2DeflateSerializeError`, `V2SerializeError`
+- Changelog
+
+### Changed
+- `DeserializeError` and `V2DeflateSerializeError` lost derived traits: `PartialEq`, `Eq`, `Clone`, `Copy`
+- Inner error type from `std::io::ErrorKind` to `std::io::Error` in types: `DeserializeError`, `V2DeflateSerializeError`, `V2SerializeError`, `IntervalLogWriterError` to support `to_string`.
```
jonhoo: I'd probably say "to support `Display`" rather than "to support `to_string`", though the two are pretty much equivalent.

jonhoo:
@AnderEnder Looks like there are some test failures?

jonhoo:
Fantastic, thank you!

PR #88, opened 2020-01-27
fix ci: code format

AnderEnder:


AnderEnder:
@jonhoo, should "Minimal rust version 1.37.0" be mentioned in Changelog?

jonhoo:
Hmm... We don't actually specify a minimum Rust version anywhere on HdrHistogram, but sure, mentioning it in the changelog seems reasonable :)

PR #96, opened 2020-11-30
add support for importing histogram snapshot

davepacheco:
This is a draft PR implementing support for importing histogram snapshots.  I'm not sure where snapshots originated, but I found this in the Golang implementation, which provides a `Snapshot` type with `Export` and `Import` functions:

https://github.com/HdrHistogram/hdrhistogram-go/blob/c968ba715dcd0cedeb8e9650662acc2fe5317db7/hdr.go#L18-L25
https://github.com/HdrHistogram/hdrhistogram-go/blob/c968ba715dcd0cedeb8e9650662acc2fe5317db7/hdr.go#L431-L456

I got here because CockroachDB's `cockroach workload` tool (a client-side load generator) uses this module to maintain histograms for request latency.  It can write periodic snapshots of the histogram to a file.  The serialized snapshot look like this:

```json
{"LowestTrackableValue":100000,"HighestTrackableValue":100000000000,"SignificantFigures":1,"Counts":[0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,3,1,6,2,2,3,2,7,11,10,6,7,4,3,1,2,0,0,1,0,0,1,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0]}
```

In this PR, I've essentially just ported the snapshot import method to this Rust crate.  Is this a welcome addition?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -287,6 +287,14 @@ pub struct Histogram<T: Counter> {
     counts: Vec<T>,
 }
 
+#[allow(missing_docs)]
+pub struct HistogramSnapshot<T: Counter> {
+    pub lowest_trackable_value: u64,
+    pub highest_trackable_value: u64,
+    pub significant_figures: u8,
+    pub counts: Vec<T>,
+}
+
 /// Module containing the implementations of all `Histogram` iterators.
 pub mod iterators;
 
@@ -811,6 +819,22 @@ impl<T: Counter> Histogram<T> {
         h.auto_resize = source.auto_resize;
         h.counts.resize(source.distinct_values(), T::zero());
         h
+    }
+
+    /// Construct a `Histogram` from a snapshot.
+    pub fn new_from_snapshot(source: &HistogramSnapshot<T>) -> Result<Histogram<T>, CreationError> {
+        let mut h = Self::new_with_bounds(
+            source.lowest_trackable_value,
+            source.highest_trackable_value,
+            source.significant_figures,
+        )?;
+
+        h.counts.truncate(0);
+        for count in &source.counts {
+            h.counts.push(*count)
+        }
+        h.restat(h.counts.len());
+        Ok(h)
     }
 
     // ********************************************************************************************


marshallpierce:
This appears to be solving the same problem as hdrhistogram's [log format](https://github.com/HdrHistogram/HdrHistogram_rust/blob/master/src/serialization/interval_log/mod.rs#L1), which is standardized across [multiple](https://github.com/HdrHistogram/HdrHistogram/blob/master/src/main/java/org/HdrHistogram/HistogramLogWriter.java#L18) [implementations](https://github.com/HdrHistogram/HdrHistogramJS/blob/master/src/HistogramLogWriter.ts#L13) [already](https://github.com/HdrHistogram/HdrHistogram_c/blob/master/src/hdr_histogram_log.h#L107). If cockroachdb exports their own new format, perhaps they should just adopt the log format instead?

davepacheco:
Thanks for the pointer.  To be clear, CockroachDB is using the (official) Golang implementation's snapshot type, not inventing their own thing.  (They have wrapped it with their own struct with additional metadata.)

marshallpierce:
Gotcha. It looks like the golang impl supports [logs](https://github.com/HdrHistogram/hdrhistogram-go/blob/c968ba715dcd0cedeb8e9650662acc2fe5317db7/log_writer.go) so I'm not sure why `cockroach workload` wouldn't just use that...

jonhoo:
I'm going to go ahead and close this given there's already an officially supported log format.

PR #98, opened 2021-04-04
Bump nom to 6.1.2 to resolve bitvec->funty conflict

tilpner:
This includes https://github.com/Geal/nom/pull/1286 which is a fix for https://github.com/myrrlyn/funty/issues/3


jonhoo:
I'm not sure why we should make that change in this library? The current dependency specification is the accurate representation of the minimal required version of `nom`, so increasing seems like the wrong thing to do. If a project that uses `hdrhistogram` runs `cargo update`, they'll get the more recent version as expected — why should we force it specifically in `hdrhistogram`?

tilpner:
You're right, please excuse the noise.

PR #99, opened 2021-05-04
add Add, Sub impls that take ownership

hdevalence:
This allows writing `hist1 + hist2`; because `Add` and `Sub` take receivers by
value, it's no more or less efficient than the more verbose `let mut hist1 =
... ; hist1 += hist2; let sum = hist1`.


jonhoo (approved):
This is a great addition, thanks!

jonhoo:
Published in 7.3.0!

PR #101, opened 2021-10-19
Derived SerDe for Histogram

DamonPalovaara:
Needed Histogram to implement SerDe for my use case but it was too difficult to get it to work without forking and changing, fields need to either be pub or there has to be getters for each private field to get SerDe to work on an external struct.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,12 +37,14 @@ flate2 = { version = "1.0", optional = true }
 nom = { version = "6.0.1", optional = true }
 base64 = { version = "0.13", optional = true }
 crossbeam-channel = { version = "0.5", optional = true }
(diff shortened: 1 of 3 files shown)


jonhoo:
Hi there! This is almost certainly not what you want — take a look at the [serialization documentation](https://docs.rs/hdrhistogram/7/hdrhistogram/serialization/index.html) instead for how to solve for your use-case :)

jonhoo:
The trouble is that there are many ways to serialize and deserialize a histogram, and we don't want to dictate one as "the standard". Your best bet here is to do as the linked documentation suggests and add your own wrapper type around `Histogram` that implements `Serialize`/`Deserialize` by delegating to the specific (de-)serializer you want to use (the docs have example code too).

PR #103, opened 2021-10-24
Update CI and make it run

jonhoo:


jonhoo:
@marshallpierce Looks like there's a legitimate failure on Windows here related to `base64` — do you have any insight into what might be going wrong [here](https://github.com/HdrHistogram/HdrHistogram_rust/runs/4993971046?check_suite_focus=true#step:4:354)? Relevant code:
https://github.com/HdrHistogram/HdrHistogram_rust/blob/c21970820a99a0d27746d1efd547ce6bd34384e4/tests/interval_log.rs#L120-L121

marshallpierce:
Hmm... some shenanigans with windows line endings or something like that perhaps? 

Any recommendations on how to get a useful Windows system for debugging?

jonhoo:
I've had luck with Microsoft's [Windows development environment virtual machines](https://developer.microsoft.com/en-us/windows/downloads/virtual-machines/) in the past. They're official and free.

marshallpierce:
Ah, I used those in the bad old days when IE compatibility matters. They're 20GB now? wow...

PR #106, opened 2022-02-09
Add an auto_resize accessor to Histogram

wcassels:
It seems strange to me that this isn't already present. I would certainly find it useful!

Plus, there's already a comment on line 1870 of `src/lib.rs` instructing me to use this method :grinning: 
```//  * Indicate whether or not the histogram is capable of supporting auto-resize functionality.
//  * Note that this is an indication that enabling auto-resize by calling set_auto_resize() is
//  * allowed, and NOT that the histogram will actually auto-resize. Use is_auto_resize() to
//  * determine if the histogram is in auto-resize mode.


jonhoo (approved):
Oh, yes, this is entirely an oversight, thanks!

jonhoo:
Also, if you rebase your changes onto `master`, you should get working CI :)

wcassels:
Rebased and added a test as requested.

wcassels:
> Published as 7.5.0 🎉

Awesome, thanks!

(By the way, I think there's a small mistake in the changelog; the newest addition references PR #102 instead of #106)

jonhoo:
Oops, you're right! I fixed the link, but not the text! Could you submit a PR with the fix? No longer at my computer.

wcassels:
> Oops, you're right! I fixed the link, but not the text! Could you submit a PR with the fix? No longer at my computer.

See #107

PR #108, opened 2022-03-20
Maybe fix interval logs on Windows

jonhoo:
Seems the checked-in test file has `\r\n` line endings on Windows, and
since we consider the encoded histogram as being everything up until the
`\n`, this makes the `\r` part of the encoded, which then makes it
invalid base64!


jonhoo:
cc @marshallpierce in case you're curious where this led :)

marshallpierce:
_sheepishy closes the open tabs that were reminding me to look into this_

PR #112, opened 2022-08-11
Optimize quantile_below, 100x speedup

the8472:
Benched with `-Ctarget-cpu=znver2`

```
OLD: test quantiles_below ... bench:       3,618 ns/iter (+/- 54)
NEW: test quantiles_below ... bench:          37 ns/iter (+/- 1)
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1378,9 +1378,27 @@ impl<T: Counter> Histogram<T> {
 
         let target_index = self.index_for_or_last(value);
         // TODO use RangeInclusive when it's stable to avoid checked_add
-        let total_to_current_index = (0..target_index.checked_add(1).expect("usize overflow"))
-            .map(|i| self.count_at_index(i).expect("index is <= last_index()"))
-            .fold(0_u64, |t, v| t.saturating_add(v.as_u64()));
+        let end = target_index.checked_add(1).expect("usize overflow");
+        // count the smaller half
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -1378,9 +1378,27 @@ impl<T: Counter> Histogram<T> {
...
+            // if the total didn't saturate then any partial count shouldn't either.
+            // iter::sum optimizes better than the saturating_add fallback below
+            iter.sum::<u64>()
+        } else {
+            iter.fold(0u64, u64::saturating_add)
+        };
+        let total_to_current_index = if lower {
+            total_to_current_index
+        } else {
+            self.total_count - total_to_current_index
```
jonhoo, line 1400: I don't think this will work if `total_count` is saturated?

jonhoo (changes requested):
This is a great observation! One correctness note though.

src/lib.rs
the8472: The docs already say

> If the total count of the histogram has reached u64::max_value(), this will return inaccurate results.

So I think it should be ok. At worst you get a different wrong answer.

jonhoo: Heh, that's fair.

jonhoo (approved):

jonhoo:
Released in 7.5.1 :tada:

PR #113, opened 2022-08-31
Optimize Histogram::clone and in some cases Histogram::add

arthurprs:
Optimize Histogram::clone by using a simple derived implementation

Benchmarks for cloning an Histogram created with `Histogram::<u32>::new_with_bounds(1, 100_000, 3)`

Before
`test clone ... bench:       6,513 ns/iter (+/- 66)`
After
`test clone ... bench:         532 ns/iter (+/- 16)`

and also
Optimize Histogram::add when the histogram parameters are equivalent and the target histogram is empty

Before
`test new_from_then_add ... bench:       6,513 ns/iter (+/- 66)`
After
`test new_from_then_add ... bench:         788 ns/iter (+/- 46)`

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -251,7 +251,7 @@ const ORIGINAL_MAX: u64 = 0;
 /// = 2048 * 2^(k-1)`, which is the k-1'th bucket's end. So, we would use the previous bucket
 /// for those lower values as it has better precision.
 ///
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub struct Histogram<T: Counter> {
     auto_resize: bool,
 
@@ -1765,14 +1765,6 @@ impl<T: Counter> RestatState<T> {
 // ********************************************************************************************
 // Trait implementations
 // ********************************************************************************************
-
-impl<T: Counter> Clone for Histogram<T> {
-    fn clone(&self) -> Self {
-        let mut h = Histogram::new_from(self);
-        h += self;
-        h
-    }
-}
 
 // make it more ergonomic to add and subtract histograms
 impl<'a, T: Counter> AddAssign<&'a Histogram<T>> for Histogram<T> {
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -473,10 +473,17 @@ impl<T: Counter> Histogram<T> {
                 .map_err(|_| AdditionError::ResizeFailedUsizeTypeTooSmall)?;
         }
 
-        if self.bucket_count == source.bucket_count
+        let matching_buckets = self.bucket_count == source.bucket_count
             && self.sub_bucket_count == source.sub_bucket_count
-            && self.unit_magnitude == source.unit_magnitude
-        {
+            && self.unit_magnitude == source.unit_magnitude;
+        if matching_buckets && self.is_empty() {
```
jonhoo, line 479: I'm confused — if `is_empty()` returns `true`, then surely that means we can just set all the fields to `0`, no? In which case we don't even need `copy_from_slice`, we can just `memset` all of it to 0?

src/lib.rs
```diff
@@ -1766,14 +1773,6 @@ impl<T: Counter> RestatState<T> {
 // Trait implementations
 // ********************************************************************************************
 
-impl<T: Counter> Clone for Histogram<T> {
-    fn clone(&self) -> Self {
-        let mut h = Histogram::new_from(self);
-        h += self;
-        h
```
jonhoo, line 1773: I forget why we did this originally. @marshallpierce do you remember?

jonhoo (commented):
Thanks for the PR!

src/lib.rs
arthurprs: self is the "target" histogram, so by checking if it's all 0 we can add the "source" in the form of a mem copy. I suppose at the top of this function we could check if the target histogram is empty and do nothing as well.

jonhoo: Ah, yes, had my references backwards. I think adding a check for if the target is empty to optimize that path makes sense :+1:

arthurprs: Done

jonhoo (approved):
Nice, thanks!

Just for future reference, it's generally best to avoid force-pushing, as it makes incremental reviews pretty painful :sweat_smile:

jonhoo:
Released in 7.5.2 :tada:

PR #125, opened 2023-08-19
Fix #124 Record iteration when recording only zeros

### LINKED ISSUE #124: Error when iterating over histogram with only zeros
If a histogram recorded only zeros, then the `iter_recorded` returns an empty iterator.

Reproduction:
```rust
use hdrhistogram::Histogram;
fn main() {
    let mut h = Histogram::<u64>::new(1).unwrap();
    h += 0;
    // Uncomment and suddenly the iterator returns 2 elements
    // h += 1;
    println!("Histogram: {h:?}");
    println!("Zeros: {}", h.count_at(0));
    // I expect that to return one element - 0 with count 1, or two if uncommented above
    for d in h.iter_recorded() {
        println!("{d:?}");
    }
}
```

It looks like the problem is with the iterator, as `count_at` returns the right number.

Carreau:
Fixes #124

As pointed in the comment in the corresponding issues, we want to ensure we trigger the picker logic at least once to keep the invariant of the loop, that is to say that the picking logic has been executed for last_picked_index, which is not the case for zero when there are only zeros.

I added a test that collect the record and check for their count.

Would it be better to check for the following:

```rust
assert_eq!(h.iter_recorded().collect::<Vec<_>>(), vec![IterationValue::<u64>::new(0, 1.0, 1.0, 1, 1)]);
```

It is easy to assume that `h.iter_recorded().collect::<Vec<_>>().len() == h.len()` and simplify the added test, and I'm unsure how to convey that we _really_ want to collect and check the length of all the record.

### DIFF
--- a/tests/histogram.rs
+++ b/tests/histogram.rs
@@ -565,5 +565,5 @@ fn subtract_underflow_guarded_by_per_value_count_check() {
 fn recorded_only_zeros() {
     let mut h = Histogram::<u64>::new(1).unwrap();
     h += 0;
-    assert_eq!(h.iter_recorded().collect::<Vec<_>>().len(), 1);
+    assert_eq!(h.iter_recorded().count(), 1);
 }
--- a/tests/histogram.rs
+++ b/tests/histogram.rs
@@ -560,3 +560,10 @@ fn subtract_underflow_guarded_by_per_value_count_check() {
         h.subtract(h2).unwrap_err()
     );
 }
+
+#[test]
+fn recorded_only_zeros() {
+    let mut h = Histogram::<u64>::new(1).unwrap();
+    h += 0;
+    assert_eq!(h.iter_recorded().collect::<Vec<_>>().len(), 1);
+}
--- a/src/iterators/mod.rs
+++ b/src/iterators/mod.rs
@@ -74,7 +74,7 @@ pub struct HistogramIterator<'a, T: 'a + Counter, P: PickyIterator<T>> {
     count_since_last_iteration: u64,
     count_at_index: T,
     current_index: usize,
-    last_picked_index: usize,
+    last_picked_index: Option<usize>,
     max_value_index: usize,
     fresh: bool,
     ended: bool,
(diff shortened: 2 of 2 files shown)


Carreau:
I would have preferred a fix that make the initial value of `last_picked_index` and invalid index, being a usize it can't be negative, and I don't see a performant and api compatible way to make it something else.

jonhoo:
Thanks for the thorough explanation! I think you're right that `more` is not to blame (and I agree with you it should probably be renamed one day). I _think_ the way to go here is to make `last_picked_index` an `Option<usize>`. That way you can explicitly indicate that it hasn't been set yet. It's not a part of the _public_ API, so you don't have to worry about breaking backwards compatibility. What's also nice is that once you change the type of that value in the `struct HistogramIterator` type definition, the compiler will guide you to all the places where the code has to be updated to reflect this change!

Carreau:
>  It's not a part of the public API

Oh, yeah, thanks I thought about it, but still sometime forget that structs fields are private by default, and wasn't sure about the performance implications.

Carreau:
Ubuntu beta failed, I can't rerun a single test, so closing and reopening.

Carreau:
Going through my list of opened PRs, is there anything more I can do to help here ?

Thanks !

tests/histogram.rs
```diff
@@ -560,3 +560,10 @@ fn subtract_underflow_guarded_by_per_value_count_check() {
         h.subtract(h2).unwrap_err()
     );
 }
+
+#[test]
+fn recorded_only_zeros() {
+    let mut h = Histogram::<u64>::new(1).unwrap();
+    h += 0;
+    assert_eq!(h.iter_recorded().collect::<Vec<_>>().len(), 1);
```
jonhoo, line 568:
```suggestion
    assert_eq!(h.iter_recorded().count(), 1);
```

jonhoo (approved):
Nope, this looks good, thank you!

jonhoo:
Released in 7.5.3 :tada:

tests/histogram.rs
Carreau: No, this is incorrect, the bug was that the iterator had the wrong number of values, not that the count was incorrect.

Carreau: This simplification does not test the fix as `.count()` was correct before this PR, but the iteration was skipping items. See #130.

PR #126, opened 2023-08-19
Explain why more() returns always `false` in the record picker.

Carreau:
I think it looks like a mistake to always return false, while it is correct. This is dues to the fact that `more()` should really be called something like `more_with_zero_count()`.


jonhoo (approved):
Thanks for taking the time to add this, it's almost certainly going to be helpful to some poor soul (maybe even me) in the future!

PR #129, opened 2023-11-06
base64 0.21, clap 4

serprex:


jonhoo (approved):
This looks great, thanks for taking the time!

jonhoo:
(rebased from `main` which may make updating this a little annoying — sorry about that!)

Looks like rustfmt isn't quite happy with the change?

jonhoo:
oof, looks like clap has [bumped its MSRV](https://github.com/clap-rs/clap/commit/d30d94f51c3e10fed96d0a555a0f0c5a2b3a5824) to 1.70. I think the way to go about this is to do only update the _lockfile_ to 4.3.24 with

```console
git checkout main Cargo.lock
cargo update -p 'clap@4.3.24'
```

(I _think_ that's the syntax at least)

Then, update `.github/workflows/check.yml` to have the MSRV be 1.64.0 (which 4.3.24 supported iiuc).

This way, consumers will get the newest version of clap, but we also ensure that there is _some_ version of the dependencies that still build for a relatively-old version of Rust for those stuck in the past.

jonhoo:
Thank you!

jonhoo:
Released in 7.5.4 :tada:

PR #130, opened 2023-11-11
Fix incorrect test simplification.

Carreau:
In https://github.com/HdrHistogram/HdrHistogram_rust/pull/125#discussion_r1390185430 a simplification was made without discussion:

```diff
-     h.iter_recorded().collect::<Vec<_>>().len()
+     h.iter_recorded().count()
```

This is incorrect as what this test actually tests was that the iterator has the correct length. In a perfect world both should be identical but the PR fix a case where this was not the case as some items would be skipped over.

In this I revert the change, add a comment and a second (redundant) check to make sure this is not re-simplified later.

### DIFF
--- a/tests/histogram.rs
+++ b/tests/histogram.rs
@@ -565,5 +565,12 @@ fn subtract_underflow_guarded_by_per_value_count_check() {
 fn recorded_only_zeros() {
     let mut h = Histogram::<u64>::new(1).unwrap();
     h += 0;
-    assert_eq!(h.iter_recorded().count(), 1);
+    // Do not simplify .collect::<Vec<_>>().len() to .count()
+    // This check a bug where .count() and .collect::<Vec<_>>().len()
+    // are different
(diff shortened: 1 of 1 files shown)


Carreau:
(I don't think this need a release as this is just a test fix)

jonhoo:
I don't follow this — `Iterator::count()` consumes the entire iterator and counts how many times it yields `Some`. `.collect()` into a `Vec` will also consume all the `Some`s, and it's `Vec::len` will then also hold that same value. So these should _always_ be equal. Can you expand more on how they could be different?

jonhoo:
As in, this is invoking `Iterator::count`, not the `count` method on the histogram/bucket. So it should test exactly the right thing (and the same as `.collect().len()`).

Carreau:
Hum, maybe I was not awake enough this morning. I'll recheck later when I'm
back in front of my computer.

Carreau:
I think you are right, I miss read how you simplified and messed up my git command while testing locally if backporting the test before the fix was passing or failing. Ignore me.

PR #133, opened 2025-01-07
Port to Winnow

epage:
I want to be clear, just because I did this, there is no pressure to adopt it. I did this to explore how people use nom to see if there are areas of improvement for Winnow. Turns out, this library's use of nom is pretty straightforward.

Winnow was originally forked from nom to unblock improvements on `toml_edit`.  For more context on Winnow, see
- [High-level comparison with nom](https://docs.rs/winnow/latest/winnow/_topic/why/index.html#nom)
- [Key API differences](https://docs.rs/winnow/latest/winnow/_topic/nom/index.html#api-differences)
- [Origin story](https://epage.github.io/blog/2023/02/winnow-toml-edit-combine-nom/)  including [another take on differences from nom](https://epage.github.io/blog/2023/02/winnow-toml-edit-combine-nom/#winnow) and then [more differences](https://epage.github.io/blog/2023/07/winnow-0-5-the-fastest-rust-parser-combinator-library/)

I see this has an MSRV of 1.64.  I used an older patch release to satisfy that.  Winnow's latest MSRV is 1.65.

This does have a minor breaking change in that the feature `nom` was removed in favor of `winnow`.  As these features are coming from optional dependencies, I'm assuming they are not intended to be features and you do not offer compatibility guarentees on them.


epage:
Due to `-sys` crates that don't build on my machine, I was unable to test this and am relying on CI.

jonhoo:
Thanks! I'm not looking to move away from `nom` right now — if anything I'd probably drop the dependency on a parser entirely since the parsing needs of this crate are pretty minimal in the first place.

Issue #114, opened 2022-09-12
Feature request: an unrecord() function


noguxun:
Can we implemented an interface to remove what we have Histogram::record() ? 
Like this decrement interface in another histgram crate
https://docs.rs/histogram/0.6.9/histogram/struct.Histogram.html#method.decrement

jonhoo:
Ah, interesting, yeah, we don't get that "for free" like the Go version does because they use signed integers for the counts:
https://github.com/HdrHistogram/hdrhistogram-go/blob/494271c4c016b36c8cee88480288f33b419cf7b0/hdr.go#L306-L309

I think the best way to go about this would be to add a `sub: bool` argument to `record_n_inner` here:
https://github.com/HdrHistogram/HdrHistogram_rust/blob/7c4a5fa1771d361aa728fc2f8d4d519e8ce11ef8/src/lib.rs#L868

And if it's `true` do `saturating_sub` instead of `saturating_add` in the three places that's called in that function. Then, add two interface methods `unrecord(value: u64)` and `unrecord_n(value: u64, n: T)` that call `record_n_inner` with the arg being set to `true`.

I'd be curious on the impact of this change on the benchmarks. It could be that it's worth marking `record_n_inner` as `#[inline]` so that `record` can remain optimized for the addition case without the extra branch, given how it's often used in hot loops. @marshallpierce may also have thoughts about this.

bogzbonny:
This feature is important to me... is there willingness to integrate it in if I PR this feature - should be able to figure it out based off of @jonhoo 's description. 

The only alternative I can really currently see (if I'd like to use this library and I do!!) is to create a new histogram with the entries I desire to remove then to subtract it from the original histogram... a cumbersome and seemingly inefficient way to do it.

jonhoo:
If you have time to write up a PR, then yes, I'd love to take a look!

PR #134, opened 2025-05-15
Unrecord

bogzbonny:
- added in basic unrecord functionality as per discussions in #114 
 - cargo update, and clippy fixes
 - probably should write some tests


examples/cli.rs
```diff
@@ -228,7 +228,7 @@ fn quantiles<R: BufRead, W: Write>(
 // Normally I frown on excessive use of From as it's too "magic", but in the limited confines of
 // subcommands, the convenience seems worth it.
 #[derive(Debug)]
-enum CliError {
+pub enum CliError {
```
jonhoo, line 231: I'm not sure I follow why this was made `pub`?

src/errors/mod.rs
```diff
@@ -7,7 +7,7 @@ use std::fmt;
 pub enum CreationError {
     /// Lowest discernible value must be >= 1.
     LowIsZero,
-    /// Lowest discernible value must be <= `u64::max_value() / 2` because the highest value is
+    /// Lowest discernible value must be <= \x60u64::MAX / 2\x60 because the highest value is
```
jonhoo, line 10:
```suggestion
    /// Lowest discernible value must be <= `u64::MAX / 2` because the highest value is
```

jonhoo: You've used this `\x60` version in a few places now — could you change all of them back to backtick please?

src/lib.rs
```diff
@@ -850,6 +850,14 @@ impl<T: Counter> Histogram<T> {
         self.record_n(value, T::one())
     }
 
+    /// Unrecord `value` in the histogram, removing from the value's current count.
```
jonhoo, line 853: I'd prefer to do this change in a PR that's separate from all the clippy/lint fixes. Could you split them up please?

src/lib.rs
```diff
@@ -879,12 +906,34 @@ impl<T: Counter> Histogram<T> {
...
+    /// from the resulting histogram without warning. Since the values are clamped, the histogram
+    /// will also not be resized to accomodate the value, even if auto-resize is enabled.
+    pub fn saturating_unrecord_n(&mut self, value: u64, count: T) {
+        self.record_n_inner(true, value, count, true).unwrap()
     }
 
-    fn record_n_inner(&mut self, mut value: u64, count: T, clamp: bool) -> Result<(), RecordError> {
+    fn record_n_inner(
+        &mut self,
+        sub: bool,
```
jonhoo, line 926: I think I'd prefer that we add
```rust
enum Op {
  Add,
  Sub,
}
```

as it will make the callsites a lot easier to vet. For example, it'll read:
```rust
self.record_n_inner(Op::Add, value, count, true)
```
instead of
```rust
self.record_n_inner(false, value, count, true)
```

src/serialization/interval_log/mod.rs
```diff
@@ -91,20 +91,20 @@
 //! ends up setting its internal version of StartTime.
 //!
 //! - Neither StartTime nor BaseTime are present: interval timestamps are interpreted as seconds
-//! since the epoch. The first interval's timestamp is stored to the StartTime field.
+//!   since the epoch. The first interval's timestamp is stored to the StartTime field.
```
jonhoo, line 94: Why did the indentation change for all of these lines?

jonhoo (changes requested):
Thank you! My guess is that CI will also complain about a change to MSRV, but a slight bump is probably warranted anyway. Main point is that I'd like to split out the lint fixes from the unrecord feature such that they're in separate PRs please.

tzachar:
Guys, is there a way to move this forward?

src/serialization/interval_log/mod.rs
bogzbonny: pretty sure this is just rustfmt

PR #138, opened 2026-07-02
feat: single-pass value_at_percentiles / values_at_quantiles batch API (7.2x)

fcostaoliveira:
## Summary

Add single-pass batch queries:

- `values_at_quantiles(&[f64]) -> Vec<u64>`
- `value_at_percentiles(&[f64]) -> Vec<u64>`

They resolve N quantiles/percentiles in **one** scan over `counts[]` instead of N separate
`value_at_quantile` calls. Requested quantiles are resolved in ascending-target order; the next
target is hoisted into a local so the hot loop stays a tight
`total += counts[i]; if total >= next_target` (the same shape as the singular scan) — per-crossing
bookkeeping runs only when a threshold is actually reached. Results are returned in input order.

## Benchmark

Getting all 7 of {50,75,90,95,99,99.9,99.99}, single core (Intel Granite Rapids):

| | calls/sec | µs/call |
|-|-----------|---------|
| 7× `value_at_percentile` | 24,896 | 40.2 |
| `value_at_percentiles` (this PR) | **178,326** | **5.6** |

**+616% (7.2×).**

## Correctness

- Results byte-identical to per-quantile `value_at_quantile` / `value_at_percentile`, verified for
  both sorted and unsorted inputs including the `0.0` / `100.0` edges.
- `record` and the singular query paths are unchanged.

Happy to add doc examples and a criterion/`#[bench]` entry if you'd like.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1409,6 +1409,9 @@ impl<T: Counter> Histogram<T> {
 
             while total_to_current_index < target {
                 let Some((i, count)) = counts.next() else {
+                    // target is clamped to total_count, so last bin must make this false
+                    // _except_ in the case where the histogram is empty
+                    assert_eq!(self.total_count, 0);
                     return Some(0);
                 };
                 at_count_i = i;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1323,6 +1323,9 @@ impl<T: Counter> Histogram<T> {
     ///
     /// This is simply `value_at_quantile` multiplied by 100.0. For best floating-point precision,
     /// use `value_at_quantile` directly.
+    ///
+    /// If you are trying to compute multiple percentiles, prefer
+    /// [`Histogram::value_at_percentiles`].
     pub fn value_at_percentile(&self, percentile: f64) -> u64 {
         self.value_at_quantile(percentile / 100.0)
     }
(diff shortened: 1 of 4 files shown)


fcostaoliveira:
Updated to address review: renamed `values_at_quantiles` → `value_at_quantiles` for consistency with the `value_at_*` pair; added `#[must_use]` on both; documented the empty-histogram/clamp/q==0 behavior; and **added tests** asserting the batch equals per-item `value_at_quantile`/`value_at_percentile` across unsorted + duplicate inputs, q==0.0/1.0, q>1.0, and an empty histogram. `cargo fmt --check` clean; the new code has no clippy warnings.

jonhoo:
I took the liberty to update this directly with a zero-allocation version that instead just leaves it to the caller to ensure that the values are in sorted order, and simply truncates the return if they are not!

jonhoo:
Note that this also bumps MSRV for 1.82 (from 1.64) since we make use of `use<>` syntax.

src/lib.rs
```diff
@@ -1369,6 +1374,79 @@ impl<T: Counter> Histogram<T> {
...
+        I: IntoIterator<Item = f64>,
+    {
+        let mut total_to_current_index: u64 = 0;
+        let mut quantiles = quantiles.into_iter();
+        let mut counts = self.counts.iter().enumerate();
+        let mut at_count_i = 0;
+        let mut previous_quantile = None;
+        std::iter::from_fn(move || {
+            let quantile = quantiles.next()?;
+            if previous_quantile.is_some_and(|pq| pq > quantile) {
```
tidely, line 1403: Here the quantile changes, but the previous quantile stays the same, so if the input is sorted except for one element in the middle, the iterator will return `None` once, and then proceed with iteration. This is technically okay, but just a little unusual. If this is unwanted, the easiest fix would be to fuse the iterator.

jonhoo: Yeah, as you say, this is generally fine since the contract for `Iterator` is that once you return `None`, you don't really need to guarantee anything about what may come next beyond that you don't cause UB (which we don't). In fact, arguably it is _more_ useful that the iterator returns to a usable state after it returns `None` for an out-of-order quantile.

jonhoo:
Released in 7.6.0 :tada:

PR #139, opened 2026-07-02
perf: iterate counts[] in value_at_quantile to elide bounds checks

fcostaoliveira:
## Summary

The `value_at_quantile` prefix-sum scan iterated with `for i in 0..self.counts.len()` and indexed
`self.counts[i]`. Iterating the slice directly with `self.counts.iter().enumerate()` hands the
element to the loop without a per-element bounds check on the hot path. Results are unchanged.

## Benchmark

`gnr1` (Intel Granite Rapids), single core, same-session A/B:

| | before | after | Δ |
|-|--------|-------|---|
| `value_at_percentile` (Mq/s) | 0.1742 | **0.1830** | **+5.1%** |

Percentile results byte-identical (the benchmark sink is unchanged). `cargo test` green;
`cargo fmt --check` clean.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1350,11 +1350,12 @@ impl<T: Counter> Histogram<T> {
             count_at_quantile = 1;
         }
 
+        // Iterate the slice directly so the element is handed to the loop without a
+        // per-element bounds check on the hot prefix-sum path.
+        // TODO overflow
         let mut total_to_current_index: u64 = 0;
-        for i in 0..self.counts.len() {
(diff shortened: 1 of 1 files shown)


fcostaoliveira:
Superseded by #140. That PR rewrites the same `value_at_quantile` scan as a chunked skip-scan (sum a chunk, skip if it can't reach the target, walk only the crossing chunk), which iterates chunks/elements and therefore already elides the per-element bounds check this PR targeted — a strict superset. It measures +63% read (vs the ~+5% here) with a byte-identical parity test across all counter widths. Closing in favor of #140.

PR #140, opened 2026-07-03
perf: chunked skip-scan in value_at_quantile (+63% read); supersedes #139

fcostaoliveira:
## Summary

Replaces the `value_at_quantile` prefix-sum scan with a **chunked skip-scan**. The old loop added one
count and branched on the running total for *every* element — a serial dependency chain. This sums a
fixed-size chunk (`chunks_exact(8)`) at a time — a reduction with no early exit that autovectorizes —
and skips the whole chunk while its subtotal cannot reach the target; only the single crossing chunk is
walked element-by-element. Counts are non-negative, so a chunk whose subtotal cannot reach the target
contains no crossing element: the result is **identical to the linear scan for any input**. No `unsafe`.

**Supersedes and closes #139** — the chunked scan iterates chunks/elements rather than indexing, so it
already elides the per-element bounds check that #139 targeted; this is a strict superset of that win.

## Benchmark

`gnr1` (Intel Granite Rapids), single core, `--release`, same-session interleaved A/B (base `main` vs this branch), 2 rounds:

| metric | base | this branch | Δ |
|--------|-----:|------------:|---|
| `value_at_percentile` (Mq/s) | 0.1739 | **0.2836** | **+62.9%** |
| batch of 7 (7× singular, Mcalls basis) | 24.8K | **40.9K** | **+64.8%** |
| `record` (M ops/s) | 348.9 | 348.6 | flat |

Percentile results byte-identical across base and patch (`sink`/`bsink` unchanged).

## Tests

- New white-box parity test comparing the chunked scan to a linear reference across **all `Counter`
  widths** (u8/u16/u32/u64) and a quantile sweep.
- A dedicated tail-crossing test: a `sig=0` histogram (`sub_bucket_half_count == 1`) is the only geometry
  whose `counts.len()` is not a multiple of 8, so it forces a crossing element into the `remainder()`
  tail loop. (For `sig >= 1`, `sub_bucket_half_count` is a power of two ≥ 16, so `counts.len()` is always
  a multiple of 8 and the tail is unreachable.)
- `cargo test`, `cargo fmt --check`, `cargo clippy` all clean (zero new warnings); MSRV 1.64 / edition
  2018 unaffected (`chunks_exact`/`remainder` stabilized in 1.31); no new dependencies.


src/lib.rs
```diff
@@ -1356,18 +1356,41 @@ impl<T: Counter> Histogram<T> {
             count_at_quantile = 1;
         }
 
+        // Sum bins in chunks with SIMD until we detect a chunk that passes the target;
```
Dr-Emann: This comment is **somewhat** inaccurate, we just iterate in chunks, and that allows the compiler to autovectorize better

jonhoo:
Released in 7.6.0 :tada:

